PR #1, opened 2022-03-15
adds sample metadata query script

MattWellie:
Initial script addition

Standalone script to take a project ID, and produce a PED file 

Updates the external IDs from the pedigree data to CPG IDs (expected in joint call)


reanalysis/generate_pedigree.py
```diff
@@ -0,0 +1,140 @@
...
+    fetches the project pedigree from sample-metadata
+    list, one dict per participant
+    :param project:
+    :return: all content retrieved from API
+    """
+
+    return FamilyApi().get_pedigree(project=project)
+
+
+def ext_to_int_sample_map(project: str) -> Dict[str, str]:
```
illusional, line 88: This endpoint doesn't return Dict[str, str], but List[Tuple[str, str]]. It does this because a participant can have multiple samples.

test/test_generate_pedigree.py
```diff
@@ -0,0 +1,128 @@
...
+        'individual_id': 'samx',
+        'paternal_id': 'samy',
+        'maternal_id': 'samz',
+        'sex': 1,
+        'affected': 1,
+    },
+]
+
+
+@patch('reanalysis.generate_pedigree.FamilyApi.get_pedigree')
```
illusional, line 50: Could you put this in a unittest.TestCase block and test it in a github action?

illusional (changes requested):

test/test_generate_pedigree.py
illusional: Resolving this as it's addressed in: https://github.com/populationgenomics/vigilant-winner/pull/3

reanalysis/generate_pedigree.py
MattWellie: very odd, I thought I had this wrapped in `dict()` to transform the results to a dictionary.... 

Hadn't occurred to me that the mapping here wouldn't be unique. Does that mean it's possible we end up with a one to many participant external ID to internal sample ID?

```
[(person1, sample1), (person1, sample2)]
```

illusional: Yep, very plausible! Our maps for seqr already support many samples for one participant, and at the CPG a couple of projects have already encountered this.

MattWellie: yikes, ok. This will need a bit of a re-think then. The pedigree generated here is intended for use in interpreting the results within a joint call... if multiple samples are present it will be tricky to know which sample to use 

parsing a VCF header to get a set of available samples and limiting the pedigree to those would work, but retrieving the same data from an MT is not so simple.

If this is integrated into the seqr pipeline using that same runtime PED file would be the logical choice, and this script can be discarded

illusional: I haven't reviewed the seqr loading pipeline, @vladsaveliev, do we assume that a participant only has one sample?

MattWellie: @illusional just having a think about this, it should be fine to duplicate the PED file row for every possible sample per participant. When the PED file is ingested for analysis, we'd just ignore all rows where the `individual ID` is not present in the callset. Not the cleanest solution, but Slivar already does something similar, and just logs the discrepant samples to STDERR and carries on with the present samples

illusional: Yeah that sounds reasonable to me!

MattWellie: Abandoning that idea... if a trio each have 2 samples that would have to expand out to 24 rows to ensure you have every possible sample combination. That's a recipe for chaos right there

MattWellie: resurrecting this, with `itertools.product()` used to create every possible proband-mother-father sample combination obtained from the API. In realistic use cases I expect that the samples per participant in a single cohort will typically be 1, so the worst case scenario seems highly unlikely

PR #3, opened 2022-03-15
Panelapp script and tests

MattWellie:
Adds the panelapp query script, tests, relevant test data. The PanelApp script uses the `requests` package to query for panel content. The panel ID is specified using a default argument, as this analysis is purely focused on the mendeliome panel. There was an API client, but it looks like GenomicsEngland didn't make that open source...

If a date (in the past) argument is provided, we also parse the panelapp Activities to find the latest panel version at the requested time. If the genes in the latest version differ from that point in the past they are annotated describing the changes (entirely new, or altered MOI since the older panel)

also adds a pytest run as a CI action


lgruen:
> The PanelApp script uses the `requests` package to query for panel content. There was an API client, but it looks like GenomicsEngland didn't make that open source... 

I found https://github.com/matteoferla/PanelApp_Python3_client_API, but haven't looked at it in detail. Any chance you could reach out to GEL and ask whether they're planning to make the official API client available?

Of course this wouldn't block this PR at all, but might be interesting for a follow-up PR and future applications.

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,244 @@
...
+    :param url:
+    :return:
+    """
+
+    response = requests.get(url, headers={'Accept': 'application/json'})
+    response.raise_for_status()
+    return response.json()
+
+
+def get_previous_version(panel_id: str, since: datetime) -> str:
```
lgruen, line 49: This name might be a bit misleading, as this really returns the first version *after* the given time, right?

I agree with your comment that it might be better to explicitly use a panel version here, to avoid confusion.

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def get_panel_green(
+    reference_genome: Optional[str] = 'GRch38',
+    panel_id: str = '137',
+    version: Optional[str] = None,
+) -> Dict[str, Dict[str, Union[str, bool]]]:
+    """
+    Takes a panel number, and pulls all details from PanelApp
+    :param reference_genome: GRch37 or GRch38
```
lgruen, line 87: Is this deliberately spelled with a lower case `c`?

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,244 @@
...
+        if entry_date >= since:
+            return entry_version
+    return entry_version
+
+
+def get_panel_green(
+    reference_genome: Optional[str] = 'GRch38',
+    panel_id: str = '137',
+    version: Optional[str] = None,
+) -> Dict[str, Dict[str, Union[str, bool]]]:
```
lgruen, line 84: Would it make sense to define a type for this return value, which you then also could reuse in e.g. `get_panel_changes`?

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,244 @@
...
+            'changed': False,
+            'old_moi': None,
+        }
+
+    return gene_dict
+
+
+def get_panel_changes(
+    previous_version: str,
+    panel_id: str,
```
lgruen, line 148:
```suggestion
    panel_id: str,
    previous_version: str,
```
```suggestion
    previous_version: str,
    panel_id: str,
```

requirements.txt
```diff
@@ -0,0 +1,9 @@
+analysis-runner==2.14.0
```
lgruen, line 1: Do you need the `analysis-runner` dependency?

requirements.txt
```diff
@@ -0,0 +1,9 @@
...
+click==7.1.2
+cpg-hail==0.2.85.post2
```
lgruen, line 3: In most places we're now using the official `hail` (i.e. not `cpg-hail`) package. Do you actually need this dependency though?

lgruen (changes requested):

requirements.txt
MattWellie: not for this script, but the analysis-runner package is used for the hail wrapper script which will call this. The requirements file isn't strictly limited to this PR, but I could make it this way

reanalysis/query_panelapp.py
MattWellie: Yes 🤦 from PanelApp: `"GRch38": {`


This is irrelevant as the same string is compared in a lowercase form (I expect the capitalisation in PanelApp to be 'corrected' at some point), so I'll make this lower case by default

.github/workflows/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.6'
```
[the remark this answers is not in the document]
MattWellie: i was still using a conda environment locally, which can't resolve python 3.10 and hail, but with pip it works fine. I'll revise my local environment and change this

requirements.txt
MattWellie: I've updated this to hail 0.2.90, instead of the cpg version

reanalysis/query_panelapp.py
MattWellie: Yeah, this was a tricky one, as 1 minute past midnight would be 'after' a datetime with day-level granularity. Taking a start date and adding a 24hr window would also fail if there were no updates on that date. I think scrapping all of this and going with strict versioning is the best route

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,314 @@
...
+def get_gcp_blob(bucket_path: str) -> storage.blob:
+    """
+    take a GCP bucket path to a file, read into a blob object
+    :param bucket_path:
+    :return: a blob representing the data
+    """
+
+    # split the full path to get the bucket and file path
+    bucket = bucket_path.replace('gs://', '').split('/')[0]
+    path = bucket_path.replace('gs://', '').split('/', maxsplit=1)[1]
```
lgruen, line 47:
```suggestion
    bucket, path = bucket_path.replace('gs://', '').split('/', maxsplit=1)
```

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+# panelapp URL constants
+PANELAPP_ROOT = 'https://panelapp.agha.umccr.org/api/v1'
+PANEL_ROOT = f'{PANELAPP_ROOT}/panels'
+PANEL_CONTENT = f'{PANEL_ROOT}/{{panel_id}}'
+ACTIVITIES = f'{PANEL_CONTENT}/activities'
+
+
+def get_gcp_blob(bucket_path: str) -> storage.blob:
```
lgruen, line 38: This would be a good candidate for using https://github.com/drivendataorg/cloudpathlib.

requirements.txt
lgruen: If you could strip it down to the minimum set, I think that would be helpful, so we can spot whether all requirements are really needed.

lgruen (changes requested):

reanalysis/query_panelapp.py
lgruen, line 47:
```suggestion
bucket, path = bucket_path.removeprefix('gs://').split('/', maxsplit=1)
```

reanalysis/query_panelapp.py
MattWellie: removed this activities search, forcing a simpler "provide a version number, or don't do a comparison"

reanalysis/query_panelapp.py
MattWellie: I was trying to avoid creating a custom class here, as the result is immediately dumped to JSON and stored, so it would be reduced back to a dict

lgruen: I didn't mean a custom class, but just a [`GenericAlias`](https://docs.python.org/3/library/stdtypes.html#types-genericalias):
```python
PanelData = Dict[str, Dict[str, Union[str, bool]]]
```

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,233 @@
...
+    ]:
+        if gene_ensg not in previous_genes:
+            latest_content[gene_ensg]['new'] = True
+
+
+@click.command()
+@click.option('--id', 'panel_id', default='137', help='ID to use in panelapp')
+@click.option('--out', 'out_path', help='path to write resulting JSON to')
+@click.option(
+    '--previous',
```
lgruen, line 181:
```suggestion
    '--previous_version',
```

design_docs/PanelApp_interaction.md
```diff
@@ -0,0 +1,26 @@
...
+disease. The Gene list we will use is the [Mendeliome Gene Panel](https://panelapp.agha.umccr.org/panels/137/).
+
+During an analysis run, the [Query Script](../reanalysis/query_panelapp.py) is used to pull down the latest data
+for the mendeliome panel (gene ENSG, symbol, and Mode of Inheritance). This is saved as a dictionary indexed on
+the ENSG
+
+One key use case for this application is to chart the differences in gene list(s) over time. This script provides
+two key ways to do this:
+
+1. Provide a prior version date as a command line argument. If this is done, the script will parse the Mendeliome's
```
lgruen, line 22:
```suggestion
1. Provide a prior version as a command line argument. If this is done, the script will parse the Mendeliome's
```

reanalysis/query_panelapp.py
```diff
@@ -0,0 +1,233 @@
...
+        # for some reason the build is capitalised oddly in panelapp
+        # at least one entry doesn't have an ENSG annotation
+        for build, content in gene['gene_data']['ensembl_genes'].items():
+            if build.lower() == 'grch38':
+                # the ensembl version may alter over time, but will be singular
+                ensg = content[list(content.keys())[0]]['ensembl_id']
+
+        # this appears to be missing in latest panel version
+        if symbol == 'RNU12' and ensg is None:
+            ensg = 'ENSG00000276027'
```
lgruen, line 103: Do we want such specific panel fixes here?

requirements.txt
```diff
@@ -0,0 +1,9 @@
+click==7.1.2
+cloudpathlib[azure,gs]==0.7.0
+cpg-utils==2.1.3
```
lgruen, line 3:
```suggestion
cpg-utils==2.5.1
```

lgruen (approved):
Looks great! Mentioned a few nits, but please feel free to merge after considering those.

reanalysis/query_panelapp.py
MattWellie: I've never seen that before, very slick

design_docs/PanelApp_interaction.md
MattWellie: 🤦

reanalysis/query_panelapp.py
MattWellie: ideally no, I've reviewed the panel data and found a few curation fixes in order to correctly map panelApp entities to VEP annotations. I'll remove this statement, and in future we'll be reliant on PanelApp as a source of truth

PR #5, opened 2022-03-21
Hail query stage

MattWellie:
# Fixes #

  - [Issue 4](https://github.com/populationgenomics/vigilant-winner/issues/4)

## Proposed Changes ##

  - introduces Hail Query script
  - CloudPathLib usage removes strict google.cloud.storage coupling in code
  - increased test coverage - WIP still
  - includes markdown documentation of whole process, with images for category descriptions

## Remaining Issues ## 
  - Still coupled to analysis-runner via DataProc; will be worked on later, not an MVP concern

## Checklist ##

- [x] Related Issue created
- [x] Linting checks pass

## additional ##
  - changed linter settings, so that required # common lines is higher for 'similar code' check - While using dataproc to run one script, importing `logging` setup from a utils script seems prone to errors


MattWellie:
I think the categorisation unit testing here is more elaborate than it needs to be. As the test VCF only contains one variant, there is no need to create a hail Table and do an annotation lookup to add the values into the Matrix table. 

The syntax could be useful in future if we have a number of test variants, and want to apply annotation to all of them. That multi-variant approach will be required if I ever get compound-hets migrated into Hail

design_docs/Hail_Filter_and_Annotation.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Stage Details
+
+Runs within a Hail Batch, as a Hail Query runtime. For now, due to functional constraints, the process runs inside
+DataProc on GCP. We are aiming to split the HailQuery/VEP annotation stage from this infrastructure requirement soon.
+
+This stage takes a Hail MatrixTable representation of a VCF as input, applies VEP annotation, and adds provisional
+category tags. Flow chart representations of the logic for pre-and post annotation filters are shown below
+
+![Pre-Annotation Filters](images/hail_pre_filter.png)
```
lgruen, line 15: This image looks like it's been upscaled from a really low-res version?

design_docs/Hail_Filter_and_Annotation.md
```diff
@@ -0,0 +1,67 @@
...
+
+---
+
+### Input
+
+Designed for joint-called MatrixTable. This is due to current reliance on the Seqr-loader
+pipeline at the CPG, which handles the transition from raw reads to clean, joint called, [GATK Best-Practice](https://gatk.broadinstitute.org/hc/en-us/articles/360035535932-Germline-short-variant-discovery-SNPs-Indels-)
+prepared variant data.
+
+If this process is opened up in future so that VCFs or alignment data is used, this will need to be adapted, or a pre-processing step created.
```
lgruen, line 29: Yes, would be awesome to have a skippable set of stages here (at some point in the future) that can use raw reads as input.

reanalysis/reanalysis_conf.json
```diff
@@ -0,0 +1,91 @@
...
+            "metasvm",
+            "mutationtaster",
+            "phast_cons",
+            "revel",
+            "slivar_comphet",
+            "splice_ai_csq",
+            "splice_ai_delta"
+        ]
+    },
+    "hail": {
```
lgruen, line 45: We're using Hail to implement this stage, but would it make sense to rename this stage to describe its purpose instead (e.g. `filter`)?

reanalysis/query_panelapp.py
```diff
@@ -177,9 +177,28 @@ def gene_list_differences(latest_content: PanelData, previous_genes: Set[str]):
...
+def write_output_json(output_path: str, object_to_write: Any):
+    """
+    writes object to a json file
+    AnyPath provides platform abstraction
+
+    :param output_path:
+    :param object_to_write:=
+    """
+
+    logging.info('Writing output JSON file to %s', output_path)
```
lgruen, line 189:
```suggestion
    logging.info(f'Writing output JSON file to {output_path}')
```

Also in a few other places, unless you're keen on the call to optimize out the interpolation in case the logging level is filtered.

reanalysis/query_panelapp.py
```diff
@@ -177,9 +177,28 @@ def gene_list_differences(latest_content: PanelData, previous_genes: Set[str]):
...
+    out_route = AnyPath(output_path)
+
+    if out_route.exists():
+        logging.info('Output path "%s" exists, will be overwritten')
```
lgruen, line 193: Missing parameter for `%s`?

reanalysis/reanalysis_conf.json
```diff
@@ -0,0 +1,91 @@
...
+        "description": "variables for the hail operations, including CSQ sets and filter thresholds",
+        "af_semi_rare": 0.01,
+        "min_samples_to_ac_filter": 70,
+        "ac_filter_percentage": 8,
```
lgruen, line 49: Would be great to add descriptions for these parameters, and potentially also a rationale as to why they've been set to a certain value (that's something that's sorely missing e.g. from the gnomAD QC scripts). Not sure if this file is the best place for it though or whether there should be a dedicated section in the docs.

reanalysis/hail_filter_and_classify.py
```diff
@@ -0,0 +1,676 @@
...
+    :param config:
+    :return:
+    """
+
+    # count the samples in the VCF, and use to decide whether to implement
+    # 'common within this joint call' as a filter
+    # if we reach the sample threshold, filter on AC
+    if matrix_data.count_cols() >= config['min_samples_to_ac_filter']:
+        matrix_data = matrix_data.filter_rows(
+            matrix_data.info.AC <= matrix_data.info.AN // config['ac_filter_percentage']
```
lgruen, line 222: Did you mean to use integer division (`//`) here?

reanalysis/hail_filter_and_classify.py
```diff
@@ -0,0 +1,676 @@
...
+    :param matrix_data:
+    :return:
+    """
+
+    # hard filter for quality; assuming data is well normalised in pipeline
+    matrix_data = matrix_data.filter_rows(
+        (
+            matrix_data.filters.length() == 0
+        )  # clarify with FILTER='PASS' & GATK SNP tranches
+        & (hl.len(matrix_data.alleles) == 2)
```
lgruen, line 239: Might be worth documenting what we're expecting in terms of multi-allelics.

reanalysis/hail_filter_and_classify.py
```diff
@@ -0,0 +1,676 @@
...
+
+def annotate_using_vep(matrix_data: hl.MatrixTable) -> hl.MatrixTable:
+    """
+    runs VEP annotation on the MT
+    :param matrix_data:
+    :return:
+    """
+
+    # throw in a repartition here (annotate even chunks in parallel)
+    matrix_data = matrix_data.repartition(150, shuffle=False)
```
lgruen, line 253: Should the number of partitions depend on the input size instead?

lgruen (changes requested):

reanalysis/query_panelapp.py
MattWellie: I was on the fence about this... I haven't (yet) posted in a mix of info and debug statements, but one useful debug one would be periodic calls to `matrix.row_count()` to assess the relative impact of each filtering step, which can involve massive computation just for a row count so would be debug only.

The main reason for using this syntax is that pylint was complaining...

reanalysis/hail_filter_and_classify.py
MattWellie: the AC is always an int, so it's a case of float division and `<`, or int division and `<=`. This AC filter at the moment is entirely arbitrary, so I'll raise it in the progress meeting to see if there are strong feelings on if/how to implement this

reanalysis/reanalysis_conf.json
MattWellie: added a MarkDown placeholder document, but not yet filler

design_docs/Hail_Filter_and_Annotation.md
MattWellie: These were terrible screenshots from Miro that I had to resize. I've added some better ones now

reanalysis/hail_filter_and_classify.py
MattWellie: I'm having a look at this now. 

Do you know offhand whether a repartition with a `shuffle` would affect row ordering, or is it strictly to do with the data representation behind the scenes?

MattWellie: looks like a full shuffle then a re-sort: https://discuss.hail.is/t/shuffling-and-writing-a-matrixtable-appears-to-run-the-shuffle-op-twice/2202/2

reanalysis/hail_filter_and_classify.py
```diff
@@ -0,0 +1,791 @@
...
+    """
+    filter MT to rows with normalised, high quality variants
+    :param matrix_data:
+    :return:
+    """
+
+    # hard filter for quality; assuming data is well normalised in pipeline
+    matrix_data = matrix_data.filter_rows(
+        (
+            matrix_data.filters.length() == 0
```
MattWellie, line 337: @vladsaveliev  this might be a good time to revisit the filters. In a typical VCF I would expect the 'filters' to contain 'PASS', or a reason for failure. In the post-VQSR data I'm using, `PASS` is implicit (empty set of filters), and I'm currently ignoring the scenario `FILTERS=VQSR` and `AS_FilterStatus=PASS`.

I'm going to open this up to 3 passing conditions:
1. filters = empty
2. filters = PASS
3. filters = VQSR & AS_FilterStatus=PASS

Does this seem like a reasonable idea? I'm not sure where in the pipeline the Filters=PASS disappear, but it could be some point during the joint-call/recalibration steps

MattWellie: My thinking here is that with the current setup, a Singleton VCF would likely be 100% filtered out, as I would have expected PASS variants to have the `filters` as `PASS`

reanalysis/hail_filter_and_classify.py
lgruen: With `shuffle=False`, you can only combine existing partitions, i.e. you can never introduce more partitions and the resulting partitions might not all be equally sized. `shuffle=True` requires distributing the data across nodes to create even partitions. On Dataproc / Spark, in practice this means you'll have to use non-preemptible nodes, as otherwise any preemption will require the whole shuffle to restart. This shouldn't be as much of a problem with the query service anymore, as there will be dedicated non-preemptible nodes for shuffling.

lgruen: The row ordering is still only given by the row key, i.e. this isn't really related to shuffling itself.

reanalysis/hail_filter_and_classify.py
lgruen: Cool, thanks! It's very likely that I misunderstand the intent of the filtering here, but naïvely I would have thought that this would be an expression of the form `AC / AN > threshold`. But that would translate to `AC > AN * threshold` or alternatively `AN < AC / T`, but you have `AC < AN / T` here.

reanalysis/query_panelapp.py
lgruen: Yeah, I think that would even be interesting at the `info` level. `row_count` shouldn't be expensive.

reanalysis/hail_filter_and_classify.py
MattWellie: Yeah, I think this is one to revisit in the config definition. For some reason in the config I put an int, so the calculation is `AC < AN / 8` which is just stupid. 

Will switch this back to a proportion and make this a multiple

`AN/AC > threshold`

reanalysis/query_panelapp.py
MattWellie: It was definitely time-expensive in the notebook I was running, which is probably the wrong point of reference 😆

reanalysis/reanalysis_conf.json
MattWellie: Added a configuration file, but no specific scientific basis for most of the parameters

reanalysis/hail_filter_and_classify.py
MattWellie: Quick revert here, from the Hail docs, FILTERS=PASS is represented as an empty set, so filter 2. above is impossible in MT. Logic allows for empty, and VQSR/PASS, I'll see how many variants are retained this way

PR #14, opened 2022-04-01
Moi confirmation

### LINKED ISSUE #7: Detach Output from Presentation
[This code](https://github.com/populationgenomics/rare-disease/blob/initial_content/reanalysis/results_builder.py) represents the rubbish, minimal HTML presentation module for the reanalysis project. 

Instead of calling this directly from the interpretation process, serialise the results to a file, then process those results into an HTML output separately.

Motivation for this is to reduce coupling to the `presentation` module to 0, so that replacement is trivial

### LINKED ISSUE #13: Create Mode of Inheritance tests for reanalysis
Add the MOI testing component:
1. takes input
  - labelled input VCF
  - the compound hets
  - panelapp data
2. For each variant, check if the MOI is suitable WRT the PanelApp details
3. For each appropriately inherited variant, add to a final report

MattWellie:
# Fixes #

  - closes #7 
  - closes #13 

## Proposed Changes ##

  - Adds the MOI tests, and overall framework for running them
  - Integrates with the end-to-end hail Batch
  - Exports all results in JSON form

## Checklist ##

- [x] Related Issue created
- [x] Linting checks pass


MattWellie:
Requires documentation, tests, a lot really...

MattWellie:
I'll be updating this PR with expanded tests and documentation, and hopefully refining some of the code. Broadly I expect the logic to remain the same, but happy for reviewers to hold off for now until it is tidier/accompanied by a guide

design_docs/Hail_Filter_and_Label.md
```diff
@@ -2,7 +2,7 @@
 
 This stage of the workflow uses hail query to load, annotate, and apply provisional categories to variants.
 
-[hail_filter_and_categorise.py](../reanalysis/hail_filter_and_categorise.py)
+[hail_filter_and_categorise.py](../reanalysis/hail_filter_and_label.py)
```
lgruen, line 5:
```suggestion
[hail_filter_and_label.py](../reanalysis/hail_filter_and_label.py)
```

reanalysis/hail_filter_and_label.py
```diff
@@ -690,23 +696,40 @@ def informed_repartition(
...
+    tmp_2 = temporary_path + '_2'
+
     # calculate partitions, falling back to 1 partition if size is too small
     current_rows = matrix.count_rows()
     if post_annotation:
         partitions = current_rows // 200000 or 1
+    elif tiny:
+        partitions = 10
     else:
         partitions = current_rows // 100000 or 1
```
lgruen, line 714: Do we really need all these different cases?

reanalysis/hail_filter_and_label.py
```diff
@@ -690,23 +696,40 @@ def informed_repartition(
...
 
-    # repartition with the specified # partitions
-    matrix.repartition(n_partitions=partitions, shuffle=True)
+    matrix.checkpoint(tmp_2, overwrite=True)
+    hl.read_matrix_table(tmp_2, _n_partitions=partitions).write(
+        temporary_path, overwrite=True
+    )
+    matrix = hl.read_matrix_table(temporary_path)
```
lgruen, line 720: Might be worth checking whether a simple `repartition` call is really that much worse when using the service backend.

reanalysis/validate_categories.py
```diff
@@ -0,0 +1,344 @@
...
+    out_json: str,
+    panelapp: str,
+    pedigree: str,
+):
+    """
+    All VCFs in use at this point will be small
+    These have been pre-filtered to retain only a small number of classified variants
+    holding all the variants in memory should not be a challenge, no matter how large
+    the cohort; if the variant number is large, the classes should be refined
+    We expect approximately linear scaling with participants in the joint call
```
lgruen, line 251: This is a key design decision that might be worth pulling out to the docs and also worth verifying on real data :).

reanalysis/utils.py
```diff
@@ -0,0 +1,342 @@
...
+# modify how to import cache based on the system python
+v_info = sys.version_info
+if v_info.major != 3:
+    raise Exception('Come on, use Python3')
+if v_info.minor < 9:
+    from functools import lru_cache as cache
+else:
+    from functools import cache
+
+# sample: gene: variant: [partner_variant, ...]
```
lgruen, line 27: Might be worth explaining the string format of each of these (or potentially adding more detail below for `VARIANT_STRING_TEMPLATE` etc).

reanalysis/utils.py
```diff
@@ -0,0 +1,342 @@
...
+from csv import DictReader
+import json
+import re
+import sys
+
+import cyvcf2
+from cloudpathlib import AnyPath
+from cyvcf2 import Variant
+
+# modify how to import cache based on the system python
```
lgruen, line 18: I think we should just assume Python >= 3.10.

lgruen (changes requested):

reanalysis/hail_filter_and_label.py
MattWellie: No 😅 this was trying to vary the partitions to get anything to complete, but the issue looks to be memory rather than the # partitions

MattWellie:
Continuing to build up the tests, but overall logic is in a fit state for review

reanalysis/hail_filter_and_label.py
MattWellie: I've re-written this to use a simpler repartition, at the moment with/without repartitioning the only failure appears at the `collect()` step, so it's hard to judge efficacy.

reanalysis/validate_categories.py
MattWellie: I've tried to clarify my expectation here, i.e. the small memory footprint of each individual variant, and the intention that the overall variant counts be v. low (e.g. 1 per variant at the end of the process). This should be a self-fulfilling scaling... If this stage of the process contains too many variants, the solution will be to revisit the filtering stage to increase the signal-noise ratio by increasing stringency.

It is a little vague still, and remains to see more real-world usage on different callsets

reanalysis/hail_filter_and_label.py
```diff
@@ -674,48 +679,33 @@ def green_and_new_from_panelapp(
...
     https://discuss.hail.is/t/best-way-to-repartition-heavily-filtered-matrix-tables/2140
 
     :param matrix:
-    :param post_annotation:
-    :param temporary_path:
     :return: repartitioned matrix
     """
 
-    # calculate partitions, falling back to 1 partition if size is too small
+    # calculate partitions, falling back to 100 partitions if size is too small
```
lgruen, line 697: Outdated number?

reanalysis/hail_filter_and_label.py
```diff
@@ -674,48 +679,33 @@ def green_and_new_from_panelapp(
...
     current_rows = matrix.count_rows()
-    if post_annotation:
-        partitions = current_rows // 200000 or 1
-    else:
-        partitions = current_rows // 100000 or 1
+    partitions = current_rows // 200000 or 50
```
lgruen, line 699: I don't think you'd want to artificially increase the number of partitions if you only have a few rows.

```suggestion
    partitions = current_rows // 200000 or 1
```

reanalysis/hail_filter_and_label.py
```diff
@@ -674,48 +679,33 @@ def green_and_new_from_panelapp(
...
 
-    # repartition with the specified # partitions
-    matrix.repartition(n_partitions=partitions, shuffle=True)
+    if partitions > current_rows:
+        partitions = 1
```
lgruen, line 703:
```suggestion
```

reanalysis/utils.py
```diff
@@ -0,0 +1,382 @@
...
+HETALT: int = 1
+UNKNOWN: int = 2
+HOMALT: int = 3
+
+# in cyVCF2, these ints represent HOMREF, and UNKNOWN
+BAD_GENOTYPES: Set[int] = {HOMREF, UNKNOWN}
+
+
+@dataclass
+class PedPerson:
```
lgruen, line 48: I think @vladsaveliev / @illusional have already built pedigree classes or are reusing the Somalier ones?

lgruen (approved):
I wonder if Alicia Byrne would be interested in reviewing the actual classification logic, which I mostly glanced over.

PR #17, opened 2022-04-11
Family analysis

### LINKED ISSUE #15: Upgrade to Family analysis
Current implementation only allows singletons, and doesn't evaluate variant calls in a family context e.g. a monoallelic variant in a proband & unaffected parent(s) will still be selected into final summary. 

1. create a pedigree representation that links all members to their relatives, so from a variant call we can traverse up/down the family tree
2. alter MOI logic to accommodate these screening steps: before 'confirming' MOI, first check validity in other family members

- Assumption so far is that families will be max. 2 generations, but the logic shouldn't force this
- Do not allow for partial penetrance just yet... but don't make it too difficult to port in

### LINKED ISSUE #16: Partial Penetrance
Current assumption is that variant = plausibly causative

Corollary is that presence of variant(s) in unaffected participants is enough to remove variant(s) from consideration

Allow for partial penetrance, e.g.

- proband has variant of interest
- parent also has variant
- inheritance for the disease/gene/participant permits partial penetrance

under this dummy example, presence in the parent isn't enough to remove the variant from consideration, but a partial-penetrance flag should be added to the variant selection model to flag that this was the situation

Decide with clinical team (will inform information source, and how interpretation logic is implemented):

- is a variant partially penetrant
- are variants in a gene partially penetrant
- is a disease evaluated as partially penetrant
- is a family interpreted under a partially penetrant model

### LINKED ISSUE #23: Solve the case of the missing variant
Your challenge, if you choose to accept it, is to explain the lack of [THIS variant](https://seqr.populationgenomics.org.au/project/R0011_acute_care/saved_variants/variant/SV0000032_2063406931_f000183_f) in the [family report](https://test-web.populationgenomics.org.au/acute-care/reanalysis/2021-09-03/summary_output.html).

Documentation of this process will be useful when it comes to implementing an automated discrepancy hunting process.

Deets: 

1. The variant is present in the proband, and not the parents
2. Var is a category_3 in the labelled VCF from Hail (High Confidence from LOFTEE)
3. Gene is GREEN
4. Var absent from both parents
5. Gene MOI is Monoallelic

Suspects:

1. Family analysis process, are we accidentally considering family members affected? Are we considering this proband unaffected?!
2. Discrepancy between class/category in code? (probably not)

MattWellie:
# Fixes #

  - Closes #15
  - Closes #16 
  - Closes #23 (regular expression error skipped chr10 and chr20)

## Proposed Changes ##

Utilises [peddy](https://github.com/brentp/peddy) as a representation of the Pedigree file, in order to implement multi-member familial analysis; each variant only 'passes' if:
  - complete penetrance (can't currently be overridden): 
      - all affected members have the variant, and no unaffected members do
  - incomplete penetrance (can't currently be triggered):
      - all affected members MUST have the variant
      - unaffected members may have the variant

Also creates similar logic for compound-hets, meaning this check is done in two phases:
 - find possible variant partners
 - iterate through family members, checking that the above logic holds (i.e. affected = has both variants, unaffected = does not have both variants present, we do not check presence of only one variant)

This is required to move from singleton analysis to familial, but can be implemented at any time. Will still only be activated if family structures exist within the provided pedigree file.

## Checklist ##

- [x] Related Issue created
- [x] Linting checks pass


MattWellie:
I'm considering this PR on hold, while I replace the bespoke Pedigree implementation with Peddy

design_docs/MOI_Checks.md
```diff
@@ -159,3 +164,31 @@ The usage paradigm is:
 4. Where a variant passes all conditions within a filter class, a 'result' object is created
 5. The result of `.run()` is a list of valid modes of inheritance (ReportedVariant), each including details of the
 variant(s), and samples
+
+### Family Checks
+
+Instead of a tree-traverse of families, the method used in family checks here is to represent each family as a flat pool
```
illusional, line 170:
```suggestion
Instead of a tree-traverse of families, the method used in family checks is to represent each family as a flat pool
```

design_docs/MOI_Checks.md
```diff
@@ -159,3 +164,31 @@ The usage paradigm is:
...
+directional manner, e.g. when considering each candidate variant:
+
+- for a monoallelic variant, inherited with complete penetrance, we enforce a unilateral rule - every person with the
+variant must also be affected. If this rule is violated for any individual member, the variant does not pass the
+inheritance test.
+- for a biallelic variant pair, all participants with the variant pair must also be affected. If the variant considered
+is Homozygous, we permit unaffected members to be heterozygous. Similarly, for compound-het inheritance, unaffected
+participants may have either variant but not both.
+
+Code has also been written to enable Complete and Incomplete penetrance models to be used. For the Complete penetrance
```
illusional, line 181:
```suggestion
Complete and Incomplete penetrance models to be used are used in this module. For the Complete penetrance
```

illusional (commented):
I don't feel qualified to properly give this the seal, programatically your changes look fine, but I don't know semantically

reanalysis/moi_tests.py
```diff
@@ -175,41 +186,119 @@ def __init__(
...
+
+        :param sample_id:
+        :param called_variants_1: called samples for variant 1
+        :param called_variants_2: called samples for variant 2
+        :param complete_penetrance: if True, (force affected==has variant call)
+        :return:
+        """
+
+        # initial value
+        variant_passing: bool = True
```
lgruen, line 280: This never gets updated, so maybe just return `True` in the end?

lgruen (approved):

reanalysis/moi_tests.py
MattWellie: Good spot, removed in latest commit 👍

PR #28, opened 2022-05-09
Drop cpg-pipes dependency

vladsaveliev:
# Fixes

Avoiding importing `cpg_pipes`.

## Proposed Changes

1. Requires `python_command` and `reference_path` to be in `cpg_utils.hail_batch`, added in this PR: https://github.com/populationgenomics/cpg-utils/pull/32
2. Moved command to run VEP on a VCF into a separate module `vep/jobs.py`.
3. Moved function that collects VEP results into a Hail Table into a separate module `vep/query.py`, to be submitted with `python_command`.
4. Original `annotation` module only has the `apply_annotations` function, to be submitted with  `python_command`.

## Checklist

- [ ] Related Issue created
- [x] Tests covering new change
- [x] Linting checks pass

Issue #9, opened 2022-03-30
Isolate Annotation from Seqr loading pipeline


MattWellie:
Currently the annotation step (within Dataproc) is not the sole source of annotations

As the process begins from a MT instead of a VCF, some of the annotations on that original MT are carried over instead of being updated. These are: 
- GnomAD/ExAC frequencies
- CADD
- REVEL
- SpliceAI
- Clinvar scores

The source for these annotations:
https://github.com/populationgenomics/hail-elasticsearch-pipelines/tree/main/download_and_create_reference_datasets/v02
https://github.com/populationgenomics/hail-elasticsearch-pipelines/blob/main/download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py#L39-L48

The annotations are available as a single Hail Table, which is used to annotate the MatrixTable as a join.

For now - make sure the Annotation stage can supply these annotations, with as little dependence on external libraries as possible

Future - obtain all annotations from a single source, so that this bespoke preparation isn't required?

MattWellie:
Plan with @vladsaveliev:
- AIP to take 'ownership' of the VCF annotation component ([methods present here](https://github.com/populationgenomics/production-pipelines/blob/initial/cpg_pipes/jobs/vep.py))
- Pipelines will import annotation function from here

This means:
- AIP will be dependent on cpg_utils, but not cpg_pipes, so no circular dependencies exist
- AIP doesn't require the stage wrapper framework to use the annotation process, but methods are available as stage logic for other processes
- Will remain dependent on preparation of input files for annotation, as currently defined in hail-elasticsearch-pipelines
  - @tiboloic has some thoughts on using the Gnomad-generated annotation bundles directly, which is already in use elsewhere in the CPG. Requires checks on size of data, genome build, availability, and frequency of regeneration
- Should probably generate a package wrapper for AIP once first draft is polished, so it can be imported from pypi by other processes

Next: 
- Port over the annotation function(s) and check that they work outside of production pipelines
  - will require generation of a pure VCF from the ACG cohort MT for testing

MattWellie:
Note - the current logic mixes 3 things:

1. annotating a VCF in chunks with VEP
2. applying those VEP annotations to a MT
3. applying VQSR annotations to a MT

Instead this will focus on parts 1 & 2, with any references to VQSR removed

tiboloic:
[...]

Issue #6, opened 2022-03-21
Remove Dataproc dependency


MattWellie:
Dataproc is a scaleable GCP service for running Spark applications. We are migrating our use of Hail from a dedicated PySpark back-end to a Hail Query service back-end.

Remove the current Dataproc runtime dependency from the application, making it cloud-portable

Note - Hail team haven't reliably run VEP without Dataproc yet, this may need solving first

https://github.com/hail-is/hail/blob/f9bcf97a867029d48886f3284784f7392f2f89f9/hail/python/hailtop/hailctl/dataproc/resources/vep-GRCh37.sh

https://github.com/hail-is/hail/pull/10454

PR #29, opened 2022-05-13
alternate rebase of annotation code

MattWellie:
# Fixes

  - solves #6 
  - solves #9 

## Proposed Changes

  - @vladsaveliev 's annotation implementation from the core CPG pipeline
  - Alternate rebase of #28, directly only main instead of an outdated HTML branch
  -

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/vep/query.py
```diff
@@ -0,0 +1,62 @@
...
+
+def vep_json_to_ht(vep_results_paths, out_path):
+    """
+    Parse results from VEP with annotations formatted in JSON,
+    and write into a Hail Table.
+    """
+    # Defining schema inside the function, so we can submit
+    # the function to the Batch Job:
+    schema = hl.dtype(
+        'struct{assembly_name:str,allele_string:str,ancestral:str,'
```
vladsaveliev, line 16: [nitpicking] I'd probably keep schema in the original format, and transform it in runtime. The reason is that I stole it from https://github.com/hail-is/hail/blob/cc0a051740f4de08408e6a2094ffcb1c3158ee9c/hail/python/hailtop/hailctl/hdinsight/resources/vep-GRCh38.json#L19, so it would be just a lit bit easier to keep in sync

reanalysis/annotation.py
```diff
@@ -0,0 +1,176 @@
...
+    )
+
+    ref_ht_path = ref_ht_path or reference_path(
+        'seqr/v0-1/combined_reference_data_grch38-2.0.4.ht/'
+    )
+
+    clinvar_ht_path = clinvar_ht_path or reference_path('seqr/v0-1/clinvar.GRCh38.ht/')
+
+    assert AnyPath(str(clinvar_ht_path)).exists(), f'{clinvar_ht_path} unavailable'
+    assert AnyPath(str(ref_ht_path)).exists(), f'{ref_ht_path} unavailable'
```
vladsaveliev, line 59:
```suggestion
    assert AnyPath(clinvar_ht_path).exists(), f'{clinvar_ht_path} unavailable'
    assert AnyPath(ref_ht_path).exists(), f'{ref_ht_path} unavailable'
```

vladsaveliev (approved):

reanalysis/vep/query.py
MattWellie: Yeah, I swapped it out to avoid doing the string manipulation at runtime (actually in response to the JSON parsing error which was really just a file truncation), but the replace operations are so trivial in the context of the whole process that I'll revert it to the original.

PR #33, opened 2022-05-16
De novo

### LINKED ISSUE #24: Split Category 4
Currently Cat. 4 is designed to consider high impact in silico variants as well as de novo. This category should be split in the following way:

- in silico classification moves to a new category 'Support', to better reflect how it is being utilised
- the assigned Category4 now contains only a light threshold, but is passed exclusively to a de novo MOI checker

How will this work?

- Category_4_only check will be modified to check a new flag
- If a variant has a Category4 flag, it will be checked for de novo inheritance only. If no results are found, the flag will be stripped off

How will this be substantially different from the current Monoallelic check? It won't be... yet, but de novo and monoallelic will differ in terms of permitted penetrance.

i.e. logic can be adjusted to allow clinvar pathogenic variants to pass despite inheritance from parents, but under all circumstances parental inheritance is disqualifying for a de novo check

MattWellie:
# Summary

This PR introduces `de novo` variants as a separate category. In the original Miro-board specification, 'Category 4' contained both strong _in silico_ results, and _de novo_ results, but only the _in silico_ part was implemented. This PR splits the _in silico_ results into a lower priority category (`support`) and category 4 now becomes entirely dedicated to _de novo_ results. 

The _de novo_ logic is implemented entirely in Hail, using the [built-in method](https://hail.is/docs/0.2/methods/genetics.html#hail.methods.de_novo). This validates possible de novo calls using a number of different elements, including read depth, allelic balance, and prior knowledge of population frequency. This content has been updated in the relevant AIP README.

The outcome of the Hail analysis is a new field in the MatrixTable: `Category4`. Instead of containing a simple boolean flag like the other categories, this contains a String; comma-delimited list of all samples with a confirmed _de novo_ hit. e.g. `Category4='Sam1,Sam34'`. This string defaults to `'missing'` when no _de novo_ events are found.

When parsing this field downstream, we split this field into the component sample IDs. A _de novo_ call is considered confirmed if a sample has a Het. genotype call, and has their Sample ID recorded in this field.

There are numerous changes to accommodate this alteration, e.g. all prior references to category 4 have been bumped to `category_support`, driver memory has been boosted to prevent an OOM failure scenario, default pedigree file format is now Plink (required by Hail), and when using a _de novo_ flag for a primary or second hit (comp-het) for a sample there is an explicit check that:

1. The variant is flagged as _de novo_
2. This particular sample had this variant as a confirmed _de novo_

During the MOI tests, category 1, 2 or 3 flags are accepted automatically. It's only cat. 4 which has this extra scrutiny.

## Fixes

  - closes #24 

## Proposed Changes

  - Uses Hail's implementation of [_de novo_ finder](https://github.com/ksamocha/de_novo_scripts/blob/master/de_novo_finder_3.py) to make a new category definition
  - The _in silico_ category is now renamed `category_Support`, to better reflect its purpose

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/validate_categories.py
```diff
@@ -207,19 +220,16 @@ def clean_initial_results(
...
-            clean_results[each_instance.sample] = {}
-
         # get an existing object, or use the current one
-        variant_object = clean_results.setdefault(each_instance.sample, {}).setdefault(
+        variant_object = clean_results[each_instance.sample].setdefault(
             var_uid, each_instance
         )
 
         # combine any possible reasons, and add
         clean_results[each_instance.sample][
```
cassimons, line 229: Can this be simplified to:
`variant_object.reasons.update(each_instance.reasons)`

cassimons (commented):
Other than the logic around MOI escape that we discussed this looks good to me (for what that is worth).

reanalysis/validate_categories.py
MattWellie: Aye, much cleaner!

MattWellie: Changed this whole section to something simpler - if something is already at that index, update. If it's empty, populate

KatalinaBobowik:
This is looking really great @MattWellie! As a very general, overarching comment, I think the code is really clean and easy to follow 👍

Issue #19, opened 2022-04-13
Create Comparison/Review process


MattWellie:
We would like a way to take external results and automatically review against the algorithm

For this we are assuming that there is a results set produced by this application, and from an external source, both based on the same raw data. i.e. manual curation results Seqr, and a result JSON from an AIP run on the same joint call

- Parse  flagged variant output summary from Seqr as 'truth' input
- Use the dumped results JSON from the MOI review phase as 'internal' results
- Identify common results between the two sets
- Identify discrepancies

Discrepancy review - where a variant was flagged externally and missed internally, aim to identify why. This should be discussed during through manual review first to come up with a suitable strategy, then codified. 

Variants in external data will have tags added to express how likely it is that AIP should catch this variant. Initially start with a traffic light estimate (Yes/Maybe/No). Any variants without any of these tags should be ignored for the purposes of this comparison.

Results from this review should be reviewed with a group of stakeholders to determine the outcome:
- are specific discrepancies to be treated as bugs, and logic changed as a result
- should the manual curation be more stringent/lenient
- are we happy with some discrepancies, allowing that this is designed to be a very specific implementation, and that may be too specific for some corner cases (it is not intended to entirely replace manual review, and as such some discrepancies may be fine)

PR #36, opened 2022-05-25
AIP part 1 - comparison documentation

MattWellie:
# Proposed Changes

  - Context: #19 
  - Adds documentation describing a high-ish level overview of the planned comparison process
  - Feature branch, adding to the overall comparison process

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


cassimons (commented):
This looks really good - but is it more than we need for MVP?  We absolutely need to quantitate which expected variants we missed, but at this point do we need the logic to determine why it was excluded?

MattWellie:
> This looks really good - but is it more than we need for MVP? We absolutely need to quantitate which expected variants we missed, but at this point do we need the logic to determine why it was excluded?

@cassimons it's definitely overkill for the bare bones MVP requirements! In the spirit of having a number of easily reviewed PRs, I see this as the first in a chain of:

1. Add documentation for end-2-end process #36 
2. Parse the AIP results JSON #37 
3. Parse the Seqr flagged variants TSV #38 
4. Compare those two datasets, and identify per-sample variant discrepancies
5. ... progressively more in-depth logic for understanding _why_ things were missed

For this purpose I think the first 4 are all logical blocks to be reviewed separately, and together form the strict MVP

cassimons (approved):
LGTM

PR #37, opened 2022-05-26
AIP part 2 - AIP Result parser

MattWellie:
# Fixes

  - partial progress towards #19 

## Proposed Changes

  - Takes input from a prior AIP run, and digests into a simple dictionary structure
  - No annotations need to be gathered from the relevant AIP results, as we will be revisiting the variant in the annotated MT/VCF as appropriate.
  - This will be matched up with a similar structure from the Seqr export, as a separate feature
  - 'Confidence' will be populated from Seqr data depending on tags; here we default to 'Expected'... because we already know it was found, and Expected is the most confident training tag in Seqr.

## Incidental Changes

- The method `read_json_dict_from_path` was a clone of the method `utils.read_json_from_path`. The former has been removed, and all JSON reading has been revised to use the utils version of the method (in program and tests)

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/comparison.py
```diff
@@ -0,0 +1,154 @@
...
+
+        for var in variants.values():
+            coords = var['var_data']['coords']
+            sample_dict[sample].append(
+                CommonFormatResult(
+                    coords['chrom'],
+                    int(coords['pos']),
+                    coords['ref'],
+                    coords['alt'],
+                    [Confidence.EXPECTED],
```
cassimons, line 124: It is not clear to me why you are using Expected as a default value here - but I am guessing it will become clear with the rest of the implementation.

cassimons (approved):

reanalysis/comparison.py
MattWellie: The thinking here is just that 'expected' is the highest confidence of the 3 options from seqr, and this is the default from the AIP side as 'we already got the result, so we're as confident as can be that we could do it again'. This could just as easily be an empty list...

PR #38, opened 2022-05-26
AIP part 3 -Add Seqr TSV parser

MattWellie:
# Fixes

  - Contributes towards #19 
 
## Proposed Changes

  - Adds a Parser for the TSV export (tagged variants) from Seqr
  - Seqr export contains Sample[1,2,3], but no notion of Proband? Instead of assuming that Sample1 is always the Proband, this takes in the relevant Pedigree and finds all Probands (no children, status: affected)
  - For all sample columns in a Seqr-export Row, find the proband by identifying a sample which has
      - Alt Calls > 0
      - Sample ID represents a Proband in the PED
  - Generate a simple representation of each variant, and append to a list for the Proband sample
  - Eventual outcome here should be a Dictionary per sample:

```
{
    'sample1':    [
        Variant(chr, pos, ref, alt, [Relevant, Tags]),
        Variant(chr, pos, ref, alt, [Relevant, Tags])
    ],
    'sample2': ...
}
```

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/comparison.py
```diff
@@ -128,18 +139,108 @@ def common_format_from_results(results_dict: Dict[str, Any]) -> CommonDict:
...
+                            confidence=tags,
+                        )
+                    )
+
+    logging.info(f'Variants from Seqr digest: {sample_dict}')
+
+    return sample_dict
+
+
+def find_probands(pedigree: Ped) -> List[str]:
```
cassimons, line 198: Nitpick: By definition, a family can only have a single proband and it is not always a child.  Most of the time in our cohorts this will be a distinction without a difference, but I would still rather not use this language.

reanalysis/comparison.py
```diff
@@ -128,18 +139,108 @@ def common_format_from_results(results_dict: Dict[str, Any]) -> CommonDict:
...
+                Confidence(tag)
+                for tag in entry['tags'].split('|')
+                if tag in VALID_VALUES
+            ]
+
+            # no relevant tags, not interested...
+            if len(tags) == 0:
+                continue
+
+            # seqr export currently has no notion of 'proband', so find the
```
cassimons, line 176: So the logic here is that you will get all affected children? If you are optionally taking multiple family members, why not just take all affected individuals? OR if you are expecting only one individual per family do you need a break after you append to the sample dict?

cassimons (commented):
Look good other than the question of if we need to tighten up the logic of which individual/s we are testing.

reanalysis/comparison.py
MattWellie: updated - this method is retitled `find_affected_samples`, and picks up all affected participants in the pedigree

reanalysis/comparison.py
MattWellie: logic changed up, now we gather all affected participants on each row with a non-WT, and all are assigned the described variant, e.g.

- 2 affected children, both het: the commonFormatVariant will be given to both samples
- 2 affected members, one WT & one het: variant only given to het sample

PR #40, opened 2022-05-27
AIP part 5 - Adds in VCF check

MattWellie:
# Fixes

  - Contributes towards #19 

## Proposed Changes

  - Core functionality: adds a new aspect to the AIP check - for each variant missing from the final results, check if the variants made it into the labelled VCF at all. This is the first major dividing point, used to triage variants into [Successful classification/MOI test required] and [Not classified/MT test required]
  - Normalises contig name when creating a commonVariant object
  - Normalises contig name when querying VCF to a contig in the header
  - Adds some additional gubbins, for general clarity and to remove redundancy:
    - Adds a glossary, documenting content & purpose of the input test fixtures
    - Removes one JSON file which was no longer required in unit tests
    - Alters the panelapp tests to close file readers

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/comparison.py
```diff
@@ -58,25 +59,37 @@ def __init__(
...
-
+        always normalise contig - upper case, with no CHR prefix
         :param chrom:
         :param pos:
         :param ref:
         :param alt:
         :param confidence:
         """
-        self.chr: str = chrom
+        self.chr: str = chrom.upper().strip('CHR')
```
cassimons, line 69: nit: strip('CHR') is not removing the string 'CHR', but is removing any leading or trailing instances of those characters. In this case I can not imagine it ever would do the wrong thing, but I frequently overlook this distinction and it can turn into a foot gun.

cassimons (commented):
LGTM

reanalysis/comparison.py
MattWellie: Aye, I've been stung by that before... without moving to regex `string[string.startswith(prefix) and len(prefix):]` can do it. I'll note this change when I get to this PR

PR #42, opened 2022-05-30
AIP part 7 - adds quality & category validation tests

MattWellie:
# Fixes

  - contributes to #19
  - splits a number of the methods on `main` to be more modular, i.e. the 'good quality variant' test monitored both `alleles` and `FILTERS` - this is now split into two methods, and the changes are cascaded accordingly

## Proposed Changes

  - The preceding branch contained testing up until the point where we had found variants of interest which were in the MT, and had not proceeded to the VCF (i.e. were called but not categorised)
  - Here we take each of those variants, and procedurally investigate why they were not assigned categories
    1. First isolate the variant of interest in the MT (throw an error if we obtain != 1 row)
    2. Prepare the variant in the same way we pre-process in the main workflow (move variants to new locations in the MT, split each gene onto separate rows, filter assigned consequences to match the individual gene ID)
    3. Run quality checking on the variant, confirming that it would have passed each Q test (AB ratio, no `FILTERS` assigned, `ALLELES` show that the variant is appropriately normalised, gene is `GREEN` in PanelApp) - if any of these tests fail, create a String explaining the failed test, and move onto the next variant.
    4. If the variant passes those quality filters, expose it to a modular version of each `category` test e.g. if the category is satisfied by being VEP HIGH consequence, then either pathogenic in clinvar, or validated by LOFTEE, run each element of that as a separate test. For each failing test, append a String explanation, e.g. `C2: Variant not new in PanelApp`
    5. For each variant obtain a list of all reasons for exclusion, and publish as a JSON summary for manual review

The intention here is that as the Seqr-flagged variant set expands, this can be re-run, and produce a breakdown of why individual variants didn't pass the classification criteria. These should be validated manually (i.e. confirm that this method is working as intended), and discussed to see whether we are happy with the discrepancies between manual and automated review.

Next steps: bin the results by strength of the underlying expectation in the Seqr output, so that we stratify these investigations. Higher priority investigation should be done for variants we were confident would be found by AIP.

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/comparison.py
+++ b/reanalysis/comparison.py
@@ -844,11 +844,14 @@ def main(
         green_genes=green_genes,
         new_genes=new_genes,
     )
+    if untiered:
+        logging.info(f'Untiered: {json.dumps(untiered, default=str, indent=4)}')
 
-    logging.info(f'Untiered: {json.dumps(untiered, default=str, indent=4)}')
-    logging.info(f'Missing: {json.dumps(not_present, default=str, indent=4)}')
+    if not_present:
+        logging.info(f'Missing: {json.dumps(not_present, default=str, indent=4)}')
 
     # write the output to a file as JSON
+    logging.info(f'Writing output JSON to {output}')
     with AnyPath(output).open('w') as handle:
         json.dump(untiered, handle, default=str, indent=4)
 
--- a/reanalysis/comparison.py
+++ b/reanalysis/comparison.py
@@ -752,8 +752,6 @@ def check_mt(
         reasons.extend(check_cat_3(matrix=var_mt, config=config))
         reasons.extend(check_cat_support(matrix=var_mt, config=config))
 
-        print(reasons)
-
         # log all reasons, even if the list is empty
         untiered[sample].append((variant, reasons))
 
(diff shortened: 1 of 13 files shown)


reanalysis/comparison.py
```diff
@@ -379,24 +701,86 @@ def check_mt(
...
+
+        # check for any failure reasons in the QC tests
+        reasons: list[str] = []
+
+        # run all methods relating to the quality of the variant call
+        reasons.extend(run_ac_check(var_mt, config))
+        reasons.extend(run_quality_flag_check(var_mt))
+        reasons.extend(check_variant_was_normalised(var_mt))
+        reasons.extend(filter_sample_by_ab(var_mt, sample))
+        var_mt = check_gene_is_green(matrix=var_mt, green_genes=green_genes)
```
cassimons, line 719: I know this is not part of this change, but should check_gene_is_green be called filter_to_green_genes or similar?

cassimons (approved):
LGTM

reanalysis/comparison.py
MattWellie: I think I'd like to go the other way, where the AIP method is a 'filter', and this checking version is always prefixed with 'check_', just so they're namespaced separately. This one is running a filter though, so you're right...

PR #47, opened 2022-06-15
Aip to seqr comparison [Checkpoint]

### LINKED ISSUE #19: Create Comparison/Review process

MattWellie:
# Fixes

  - closes #19 

## Proposed Changes

  - This 7 aggregates 7 mini-changes, together forming the bulk of the comparison process (AIP vs. Seqr)
  - This comparison is not 100% complete (i.e. there is no testing of variants which appear to have failed MOI tests), but changes up to this point have also altered the main methods forming the functional program
  - In the immediate future the comparison process may form a lower priority than working on SpliceAI results, multi-panel queries, or other elements... Divergent development will make these changes much harder to integrate in future, so merging here to form a 'checkpoint'

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


cassimons (approved):
All reviewed as part of mini-PRs.

PR #49, opened 2022-06-16
corrected annotation documentation

### LINKED ISSUE #48: Correctly document the Annotation process
Current Documentation has a number of errors and omissions

- READMEs still discuss how annotated data is required as an input, no longer true
- READMEs still describe the annotation as happening in a dataproc instance within the filtering script
- READMEs still describe (and show in flow diagrams) how filtering is carried out before and after annotation within the same process
- Diagrams still contain the original definition of 'category_4', incorporating both de novo and in silico scores

Solution:
Update documentation, and replace static (incorrect) images with modifiable flow diagrams (e.g. plantUML), so we can adapt on the fly

MattWellie:
# Fixes

  - closes #48

## Proposed Changes

  - Adds a separate flow diagram for each category definition, and for the correctly ordered filtering flow
  - Removes documentation of dataproc, and annotation using the `hail.vep()` method
  - Describes the annotation input and directory structure, and documents the source of the datasets

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


design_docs/Annotation.md
```diff
@@ -0,0 +1,54 @@
+# Annotation
+
+Prior to running the filtering and categorisation process, we need the variant data to be annotated; population
+frequencies, clinvar entries, and a range of predicted consequences. To do this we use a combination of VEP and
+annotation directly from Hail Tables containing annotations.
+
+This process is split into a number of steps:
+
+1. If the variant (singleton, or joint call) data is provided in MatrixTable format, we revert this back to a VCF
```
vladsaveliev, line 9: Detail, but I'd stress that it's "site-only VCF"

design_docs/Annotation.md
```diff
@@ -0,0 +1,54 @@
...
+2. We use GATK to identify intervals across the genome, so the VCF can be evenly split and annotation tasks parallelised
```
vladsaveliev, line 10: "evenly". Ideally we want dynamic batches, so we can split into regions based on amount data, and not based on pre-defined fixed intervals, so it would be much more even :) But that's down the road

vladsaveliev (approved):
Looks great!!
A couple of minor comments :)

PR #52, opened 2022-06-21
Remove hail comp hets

### LINKED ISSUE #51: Compound-Hets are ruining my day
Ref issue:  https://hail.zulipchat.com/#narrow/stream/223457-Hail-Batch-support/topic/OutOfMemoryError.20in.20ServiceBackend.2ElowerDistributedSort/near/277221590

Identification of compound-het variant pairs in AIP is currently done in Hail. Once the MatrixTable has been minimised to contain only categorised variants, I run [this method](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/1e871e388799c209ff8e4ce356a25e08052c83c7/reanalysis/hail_filter_and_label.py#L356-L416) to collect any heterozygous variant pairs present on the same gene (as identified by gene symbol).

This collect function works in a local notebook, or a Dataproc instance, but is having a terrible time in Hail Query... There is some memory leak during the collection (possibly due to over-zealous logging?) that results in the worker node running out of memory. Hail query doesn't currently permit manually setting higher memory allowances on worker nodes, so this is a hard problem to treat. We have been able to keep just ahead of this issue through a number of tweaks, but every time the number of variants rises (e.g. implementing a new category) we encounter new failures.

I'm sure there is a simple solution that will solve this forever, but I'm not sure what it is, and at this point the development of tweaks and failed runs caused is consuming an unjustifiable amount of time.

Solution:

- remove current comp-het checking method
- manually generate the collection of comp-het variants

Where?

- when we iterate over the VCF, we are already collecting variants by-gene
- after collecting all variants within a gene, check for comp-hets using autosome-aware collection
- see if we can acknowledge variant phase (not currently taken into account in hail) 

Proposed solution: remove the current comp-het method, and implement manually

MattWellie:
# Fixes

  - fixes #51 

## Apologies

This isn't a neatly packaged PR. It does mainly address one core change, but that inserting that ended up being noisier than expected. Phase-aware checking was an almost-free upgrade during the process, though it could have been packaged as a separate PR...

## Proposed Changes

  - removes the fragile compound-het logic from hail
  - variants are read from the VCF and cast as AbstractVariants, stored in a dictionary indexed by gene name (unchanged behaviour)
  - as AbstractVariants are created, their phase set and phased genotype (e.g. '1|0' vs '0|1') at the locus is stored
  - when running all variants in a gene, we manually check for compound hets 
      - this check is autosome/sex aware (no C-H on Y, no Male C-H on X)
      - this check is also phase aware (if one sample is het for two variants in the same gene, but the two variants are in phase, the possible comp-het is ignored)

## Actual Changes

- removed comp-het logic from hail classification
- changes a number of the type hints to python types e.g. 
  - Dict -> dict
  - Optional[str] -> str | None
- removes the test case for compound-het gathering in hail, and the associated VCF fixtures
- updates the `interpretation_runner.py` wrapper to use the new cpg_utils.config content instead of environment variables
- removes a layer of indexing from most dictionaries (previous structure was often `{"string_of_variant": variant_object}`)
- the compound-het lookup dictionary now contains the full variants (we are parsing directly from the VCF, instead of from a JSON document), so extra logic requiring variants to be looked up is thrown out
- Each MOI test is a class instance. Previously each MOI was instantiated with a few parameters, then run for each variant. The genomie-wide compound-het dictionary was provided during instantiation, but now it is supplied to `run()` - tests updated accordingly
  - reason is that MOI classes are created early, so that they only have to be created once per MOI. At this early stage, variants are not read yet, so comp-het dictionary doesn't exist
  - comp-het dictionary is now assembled once per gene, instead of holding a record of every genome-wide compound het
  - gene-only dictionary is smaller, so passing with every variant isn't a huge burden
- tests created to prove the comp-het methods function correctly on a list of variants read from a VCF
- tests created to prove that phase is correctly read from the cyvcf2 variants [see here](https://github.com/brentp/cyvcf2/issues/31), and [here (GATK's Phase grouping)](https://gatk.broadinstitute.org/hc/en-us/articles/360050354712-What-is-physical-phasing)
- test created to prove that variants which are otherwise comp-het, but in phase, are not selected by the comp-het method
- phasing implementation is sensible, but not overly complex:
    - assumes a max of 1 phase-set value assigned to any given variant
    - skips phase test on chrY
    - skips phase test on Male X
- MOI tests remain in addition to the phase check
- If two variants form a comp-het in a proband, and we can't see that the variants are physically in phase, the rest of the family are checked for the same variant pair. When doing that test, we don't check phase in the other family members. Technically that information will also be in the comp-het dict, but the probability of intragenic recombination between generations is vanishingly small... `isn't it?`
- some more stuff IDK, I got a bit carried away

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


lgruen (approved):
Not for this PR, but I was wondering whether we should try one of the probabilistic phasing tools, as we presumably often will have trios and then could get larger haplotype sections phased with high confidence? Presumably that would help a bit with the comp het determination?

MattWellie:
@lgruen that would be a fun little side project. I've not looked into statistical/imputation phasing in a long time but it could be useful for both trios and singletons

PR #57, opened 2022-06-23
allow annotation skipping (+ tests)

### LINKED ISSUE #56: Allow AIP to take correctly annotated input MT
## Issue

AIP (specifically the category annotation stage) is very hard-coded in terms of the named fields where annotations are expected to occur. The program will crash if fields have unexpected names, unexpected nesting, or are absent entirely. Because of this, the assumption is that annotation always needs to be re-generated when new data is supplied. 

As the annotation component is modelled from @vladsaveliev's CPG annotation stage, this re-annotation is redundant when data has recently been annotated by the CPG pipeline. In this situation, this would be wasteful in terms of time, money, and compute.

Also worth noting that Hail can generate a VCF from a MT, but takes its sweet time doing it. This is not a parallel operation, and transcribing a Whole Genome MatrixTable to VCF can take over an hour.

## Change

Currently input can be specified as a MT or a VCF; VCFs will be annotated, and MTs will be decomposed back into a VCF and annotated. Allow for a flag to be added such that for input MT data the `MT->VCF->Annotate->MT` loop will be skipped.

## Note

1. This is to be used sparingly, where the user is confident that the fields as annotated in the input MT exactly map to the AIP specifications.
2. When AIP runs annotation, it writes the annotated MT to a predictable per-run output path. If it finds this output MT, annotation will be skipped. This issue's behaviour could be simulated by copying the input MT to the expected output path, but MT directories contain absolutely tons of files, and copying those is not trivial either.

MattWellie:
# Fixes

  - closes #56 

## Proposed Changes

  - adds a CLI flag to signal annotation skipping
  - creates an enumeration to specifically identify different file types
  - changes the `is this a VCF` test to `what file type is this`
  - tests if an input is MT - if True, and annotation skipping flag is passed, the expected path of the annotated MT will be overwritten with the current input MT path
    - this will be read during the category assignment phase

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/interpretation_runner.py
+++ b/reanalysis/interpretation_runner.py
@@ -267,7 +267,7 @@ def handle_hail_filtering(
         f'--out_vcf {HAIL_VCF_OUT} '
     )
 
-    logging.info(f'PanelApp Command: {labelling_command}')
+    logging.info(f'Labelling Command: {labelling_command}')
     labelling_job.command(labelling_command)
     copy_common_env(labelling_job)
     return labelling_job
--- a/reanalysis/interpretation_runner.py
+++ b/reanalysis/interpretation_runner.py
@@ -15,9 +15,7 @@ re-run currently requires the deletion of previous outputs
 
 
 from typing import Any
-
 import logging
-from pathlib import Path
 import os
 import sys
 
@@ -41,7 +39,7 @@ from cpg_utils.hail_batch import (
 )
 
 import annotation
-from utils import read_json_from_path
+from utils import read_json_from_path, FileTypes, identify_file_type
 from vep.jobs import vep_jobs, SequencingType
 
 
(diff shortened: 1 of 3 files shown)


cassimons (approved):
Question: do you want/need to add any logic to test if it has the correct annotation fields in the the ht? I imagine that would be a pain to maintain comprehensively, can we add some annotation version flag to the ht that can be used to a)show the expected annotation is present and b) suggest re-annotate if some breaking change has been made to the required annotation?

MattWellie:
@cassimons good idea, i'll add an issue to investigate. The matrix `.describe()` details are accessible, but are strings rather than structured data. Possible to produce a simple pass/fail based on the fields required, but probably not so simple.

PR #60, opened 2022-06-24
Add Setup.py

MattWellie:
# Fixes

  - N/A

## Proposed Changes

  - adds a setup file
  - split requirements into main/test/pre-commit(lint), and populate in setup
  - install package inside final container
  - adds bump2version and configuration

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


setup.py
```diff
@@ -0,0 +1,56 @@
...
+        'dill==0.3.4',
+        'hail==0.2.93',
+        'Jinja2==3.0.3',
+        'pandas==1.3.5',
+        'peddy==0.4.8',
+        'requests==2.25.1',
+        'sample-metadata==4.10.0',
+        'seqr-loader==1.2.5',
+    ],
+    extras_require={
```
vladsaveliev, line 46: I like it! Will adopt `extras_require` this in my projects as well :D

`pre-commit` section seems to duplicate `requirements-lint.txt`, maybe worth reading `requirements-lint.txt` contents here, to avoid them accidentally getting out of sync?

vladsaveliev (approved):

setup.py
MattWellie: @vladsaveliev that was exactly my thinking! Then I started reading around for the best way to parse requirements into setup.py and the suggestions were to keep them separate (e.g. https://stackoverflow.com/a/16624700)

My main reason for keeping requirements-lint.txt separate is that to run linting in CI it only needs those 3 packages, but running `pip install .[lint]` would install numpy, hail, etc...

Maybe I should remove the `extras` option for lint...

MattWellie: This was a longer explanation https://stackoverflow.com/a/33685899 - I'd never used them in this separate way before

vladsaveliev: > y main reason for keeping requirements-lint.txt separate is that to run linting in CI it only needs those 3 packages, but running pip install .[lint] would install numpy, hail, etc...

That's right, `requirements-lint.txt` is certainly needed!

vladsaveliev: > Maybe I should remove the extras option for pre-commit...

I think that would make sense! 

But it's really nitpicking from me, feel free to merge as is :-)

PR #61, opened 2022-06-24
Implement Checkpoints

MattWellie:
# Fixes

  - N/A

## Benefits

After @lgruen 's feature presentation, I re-implemented checkpoints in the hail query script, to segregate & resolve blocks of filter queries [details on slack](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1656035482419579?thread_ts=1656031499.504459&cid=C018KFBCR1C). Runtime of the exact same operation was improved by ~40%.

## Proposed Changes

  - implements checkpointing in the Hail process:
    1. Checkpoint after a batch of independent filters
    2. Never run `count_rows()` prior to checkpoint
    3. flip-flop the checkpoint write-path, so we can use multiple checkpoints without trying to write to the current read-path. Method used rotates the temp path extension `.mt` <-> `.mt2`
    4. follow checkpoints with re-partitions (no way to auto-repartition AFAIK, despite checkpoint implementing a read & write)
  - alteration so that instead of separating the checkpoint, repartition, and logging, all those 3 operations are in a single method. Huge reduction in the number of times `count_rows()` is called
  - also fixes one bug on main, where unaffected samples are having variants confirmed in the report
    - for all MOI tests, only process samples as 'primary' where they are affected 
  - pre-commit removed from `setup.py` - I meant to include this in #60 but I smurfed it up

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/hail_filter_and_label.py
+++ b/reanalysis/hail_filter_and_label.py
@@ -680,22 +680,21 @@ def checkpoint_and_repartition(
     :param matrix:
     :param checkpoint_path: tmp location for checkpoint
     :param extra_logging: any additional context
-    :return: repartitioned, checkpointed matrix
+    :return: repartitioned, post-checkpoint matrix
     """
     global CHECKPOINT_EXTENSION  # pylint: disable=W0603
-    matrix = matrix.checkpoint(
-        f'{checkpoint_path}{CHECKPOINT_EXTENSION}', overwrite=True
-    )
+    checkpoint_extended = f'{checkpoint_path}{CHECKPOINT_EXTENSION}'
+    logging.info(f'Checkpointing MT to {checkpoint_extended}')
+    matrix = matrix.checkpoint(checkpoint_extended, overwrite=True)
     CHECKPOINT_EXTENSION = CHECKPOINT_FLIP_FLOP[CHECKPOINT_EXTENSION]
 
     # estimate partitions; fall back to 1 if low row count
     current_rows = matrix.count_rows()
     partitions = current_rows // 200000 or 1
 
-    log_string = f'Re-partitioning {current_rows} into {partitions} partitions '
-    if extra_logging:
-        log_string += extra_logging
-    logging.info(log_string)
+    logging.info(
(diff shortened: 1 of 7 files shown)


reanalysis/hail_filter_and_label.py
```diff
@@ -698,8 +698,14 @@ def main(
     :param config_path: path to the config json
     :param plink_file: pedigree filepath in PLINK format
     :param out_vcf: path to write the VCF out to
+    :param tmp_path: temporary checkpoint write path
     """
 
+    # when we checkpoint to a location, we are now reading from it
+    # to maintain multiple checkpoints, we rotate between two ext.
+    checkpoint_extension = ''
+    checkpoint_flip_flop = {'': '2', '2': ''}
```
lgruen: Since you're writing to a temp location anyway, I wouldn't be too worried about the flip-flopping to save storage. It can actually be kind of helpful for debugging afterwards if you have snapshots of the various stages available.

For example our temp buckets clear out any data after 7 days, so it's just a short-term extra cost to keep everything around.

reanalysis/hail_filter_and_label.py
```diff
@@ -845,11 +843,15 @@ def main(
     parser.add_argument(
         '--out_vcf', type=str, required=True, help='VCF path to export results'
     )
+    parser.add_argument(
+        '--tmp_path', type=str, required=True, help='Temp path for checkpoints'
```
lgruen, line 855: I'm not sure how much you've integrated the recent config changes, but should this be a parameter or use the `hail_batch.output_path` function (with a `tmp` category) in `cpg-utils` instead?

lgruen (approved):

reanalysis/hail_filter_and_label.py
MattWellie: I'll revert this and just increment for each checkpoint!

reanalysis/hail_filter_and_label.py
MattWellie: I am using and passing it through as an argument, but it does make more sense to just do that here!

PR #63, opened 2022-06-27
hash based subselection of members/families

### LINKED ISSUE #62: Seeded selection of Families
## Problem

AIP is currently being developed in 2 ways:

1. based on an initial design specification
2. in response to on-going analysis by curators

AIP analysis of all families going into Seqr will result in over-fitting of the algorithms to the test cases. To combat this we would like to exclude a number of families from the analysis, so that those can be naively analysed.

## Considerations

1. Excluded families should be 'random' (i.e. not manually selected)
2. Excluded families should be deterministic (i.e. as the cohort expands, when this sub-selection runs the same families should be removed from analysis as before)

## Plan! 

This sub-selection can be done by altering the input Pedigree file, which is used as a static input to the analysis (any samples not identified as affected in the Pedigree file will be skipped over in the MOI testing). The pedigree is generated in a few steps:

1. Query Sample-Metadata API for all families & members
2. [optional] Overwrite all members as singletons
3. Assemble all members into a PED/Plink format file

The proposal here would be to include a new step (between 1 & 2) to remove a subset of the families

## Specifics

`random.sample()` is unusable as repeat runs would produce inconsistent data. Even if seeded consistently, as the sample space changes (addition of more samples into the cohort), e.g.

```python
from random import sample, seed
import sys
max_range = int(sys.argv[1])
seed(1)
print(sample(range(max_range), max_range // 10))
```

```
31: [4, 18, 27]
32: [8, 4, 16]
40: [8, 36, 4, 16]
```

`hash()` is also unusable, as `hash` uses a seed which can't be set in code, meaning that either the environment has to be curated for this one process, or the results of `hash` are unreliable between runs.

A logical route here would be to perform a hashing operation on individual family IDs. That way the selection of any individual ID is independent from the population of values it occurs within:

```python
import hashlib
import sys
str_range = list(map(str, range(int(sys.argv[1]))))
threshold = 70
kept = []
for value in str_range:
    byte_value = value.encode('utf-8')
    hash_value = int(hashlib.sha1(byte_value).hexdigest(), 16) % (10 ** 8)
    if hash_value % 100 >= threshold:
        kept.append(value)
print(kept)
```
```
10: ['1', '3', '5']
15: ['1', '3', '5', '10', '12']
20: ['1', '3', '5', '10', '12', '16', '18']
```

## Problems?

I'm not 100% sure how the Hail `de novo` function will react if not all samples present in the MatrixTable are present in the provided pedigree file.

If a sample is missing from the pedigree, I need to check whether the Ped.get(SampleID).affected will fail, or if it needs to be accounted for. 

Solution for both of these points would be to sub-select the MatrixTable based on the individuals present in the pedigree

MattWellie:
# Fixes

  - closes #62 
## Proposed Changes

  - implements a mechanic in the pedigree generation script where a threshold can be provided. The threshold feeds into a hashing method, and will result in a percentage of members being removed
      - threshold == 60 means approx 40% of the PED families will be removed
  - MatrixTable is now filtered to only the samples present in the Pedigree - this should get ahead of any silly issues where the samples in the joint call and pedigree no longer match up
  - pedigree file is now read into the batch properly, instead of being parsed into temp files (3 times..)
  - subsetting syntax based off this original script https://github.com/populationgenomics/fewgenomes/blob/main/datasets/acute-care/extract_trio_vcf.py

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


MattWellie:
Test batch submitted with these changes
https://batch.hail.populationgenomics.org.au/batches/76579

cassimons (approved):
Looks great Matt. This implementation was exactly what my incoherent rambling was intending to express.

PR #74, opened 2022-06-30
Platform independence changes

### LINKED ISSUE #72: Platform independence!
Base issue for loosening the dependence on CPG-specific infrastructure:

- Scout out any hard-coding that relates to `gs://` paths, or CPG-specific paths
- Output bucket paths should be revisited, other deployments may not share the rigid `-main`, `-main-web`, `-main-tmp` bucket structure
  - Add those extensions to the config file, so that we can choose whether or not to use them - this would mean that we can still use `output_path`
  - If the extension is a missing key or empty string, we would write all the results to the same bucket/blob?
- Add Azure module to the cloudpathlib installation

### Bit more involved... 

Reduce the hard-pathing for annotation inputs, e.g. the path to the 'combined annotations HailTable' should be provided literally instead of building a path `<reference_bucket>/seqr/<HailTable_name>_<table.version>`

- This should be done such that we don't get too far away from the core CPG pipeline's annotation implementation

MattWellie:
# Fixes

  - partially resolves #72 

## Proposed Changes

Various changes to reduce hard-coded dependence on CPG infrastructure
  - we no longer assume the `bucket`, `bucket-tmp`, `bucket-web` structure. Instead, those suffixes are read from the config file and can be missing/empty strings, in which case the results will all be lumped into the same bucket
  - added the `web_suffix` and `tmp_suffix` keys to the config template
  - new method to read config and check for a 'good' (len > 0) string - if not returns None
  - no changes to the annotation component

## Incidental

Squashed some argparse/click descriptions to squash down the line count

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/utils.py
+++ b/reanalysis/utils.py
@@ -616,3 +616,14 @@ def good_string(in_string: str) -> bool:
     if isinstance(in_string, str) and len(in_string) > 0:
         return True
     return False
+
+
+def check_good_value(key: str, conf: dict[str, Any]) -> str | None:
+    """
+    check if a key from the dictionary is a good string
+    return it or None
+    """
+    conf_value = conf.get(key)
+    if good_string(conf_value):
+        return conf_value
+    return None
--- a/reanalysis/hail_filter_and_label.py
+++ b/reanalysis/hail_filter_and_label.py
@@ -14,7 +14,6 @@ Categories applied here are treated as unconfirmed
 """
 
 from typing import Any
-import json
 import logging
 import sys
 from argparse import ArgumentParser
@@ -24,6 +23,8 @@ from peddy import Ped
 
 from cloudpathlib import AnyPath
 from cpg_utils.hail_batch import init_batch, output_path
+
+from reanalysis.utils import read_json_from_path
 
 
 # set some Hail constants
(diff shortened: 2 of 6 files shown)


reanalysis/utils.py
```diff
@@ -616,3 +616,14 @@ def good_string(in_string: str) -> bool:
...
+
+def check_good_value(key: str, conf: dict[str, Any]) -> str | None:
+    """
+    check if a key from the dictionary is a good string
+    return it or None
+    """
+    conf_value = conf.get(key)
+    if good_string(conf_value):
+        return conf_value
+    return None
```
violetbrina, line 629: What is the purpose of this function over conf.get(key)? Doesn't it just return None by default?
If you're planning on adding functionality here that makes sense then.

MattWellie: There will be a template config file with this repo, with mostly blank values for the deployment-specific details. Those defaults should match the intended types, so the default value would be `""` instead of `none`. 

My thinking here is to `.get()` the value for a specific key, and return None either if the value is truly None, or if the value is an empty String. Definitely a cleaner way to do this...

violetbrina: Ah nice. Yeah sounds good.

violetbrina (approved):

MattWellie:
I've run a test, and when the extension keys are missing, output paths are generated in the regular bucket, and the '-web' suffix is added when the key is present. Seems like it'll work

Issue #70, opened 2022-06-29
Much stricter thresholds


MattWellie:
Some points from the meeting earlier, for discussion 

- Currently no specific gnomad AC count on Hemizygote MOI test
- Gnomad Homs - > 1 should be enough to dismiss any variant? > 2?
- Add gnomad exomes and genomes counts when counting the number of homs/hemi in population data?
  - I'm not super enthusiastic about this, if we're soon (?) moving to a 700k genome Gnomad v4
- AC within joint call should be much lower than 10%. Suggesting a shifting threshold so that we remain sensitive to segregation of causative variants within large families no matter the cohort size:
    - AC < 100, 5% cut-off?
    - 100 < AC < 1000, 1% cut-off?
    - AC > 1000, 0.5%?
    - Should these be specific numbers, or percentages?

pfdefazio:
Thanks @MattWellie. I can comment relating to VCGS clinical process in a rare disease context:

> Currently no specific gnomad AC count on Hemizygote MOI test

We'd typically disregard >=1 hemi as long as there are multiple hets ('multiple' is vague and up for discussion...>2?).

> Add gnomad exomes and genomes counts when counting the number of homs/hemi in population data?

Also not enthused by this.

> Gnomad Homs - > 1 should be enough to dismiss any variant? > 2?

We'd typically disregard any variant with >=1 het and >=1 hom. >=2 homs is conservative, I wouldn't go higher than that.

> AC within joint call should be much lower than 10%. Suggesting a shifting threshold so that we remain sensitive to segregation of causative variants within large families no matter the cohort size:

We use:
- AC >= 500, 1%;  OR 
- variant count needs to be >=5 (so 2% at AC = 250, 5% at AC = 100, etc) where our AC < 500

MattWellie:
I like that AC/AN threshold. I'm going to switch it up so that we'll remove variants with AC > 1%, but no filter applied to AC<=5

PR #75, opened 2022-06-30
Filter changes

MattWellie:
# Fixes

  - #70 

## Proposed Changes

  - Removed Exac checks as redundant, focus on Gnomad
  - Skips over male proband variants on X if the gnomad hemi count is high

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/hail_filter_and_label.py
```diff
@@ -576,16 +578,12 @@ def extract_annotations(matrix: hl.MatrixTable) -> hl.MatrixTable:
             exac_ac_het=hl.or_else(matrix.exac.AC_Het, MISSING_INT),
             exac_ac_hom=hl.or_else(matrix.exac.AC_Hom, MISSING_INT),
             exac_ac_hemi=hl.or_else(matrix.exac.AC_Hemi, MISSING_INT),
-            gnomad_ex_cov=hl.or_else(matrix.gnomad_exome_coverage, MISSING_FLOAT_LO),
```
cassimons, line 579: The summary says you are removing the exac checks (which makes sense), but this looks like you are leaving exac in but dropping gnomad_exome (which makes much less sense)?

MattWellie: Big ol' brainfart. 

The original MAF check was checking against gnomad genomes and exac. Should really be swapping out exac for gnomad exomes instead of removing gnomad exomes

MattWellie: updated to check check gnomad exomes and genomes, instead of gnomad genomes + exac

also updated to check hemi & homozygote ACs in both gnomad exomes & genomes

reanalysis/hail_filter_and_label.py
```diff
@@ -51,7 +49,9 @@ def filter_matrix_by_ac(
     :param ac_threshold:
     :return: reduced MatrixTable
     """
-    return matrix.filter_rows(matrix.info.AC / matrix.info.AN < ac_threshold)
+    return matrix.filter_rows(
+        (matrix.info.AC <= 5) | (matrix.info.AC / matrix.info.AN < ac_threshold)
```
cassimons, line 53: This is a nice way to deal with downscaling!

reanalysis/hail_filter_and_label.py
```diff
@@ -421,16 +421,15 @@ def filter_to_population_rare(
     matrix: hl.MatrixTable, config: dict[str, Any]
 ) -> hl.MatrixTable:
     """
-    run the rare filter, using Gnomad & exac
+    run the rare filter, using Gnomad Xxomes and Genomes
```
cassimons, line 424: Xxomes?

cassimons (approved):

reanalysis/hail_filter_and_label.py
MattWellie: Less naughty than XXXomes 😉

MattWellie:
The hail version had moved on, and the version in the requirements was now out of date. That meant that when AIP was installed inside the container it was overwriting the Driver image's version of hail, and everything broke.

This has been dealt with by including only two dependencies in the pip installation by default (as we know the driver image has all the other relevant requirements already), and the rest will be triggered by `pip install .[full]`.

Interesting that the old Hail JAR files just aren't accessible though, could have implications elsewhere.

PR #91, opened 2022-08-09
Remove second de novo

### LINKED ISSUE #90: Remove home-brewed de novo logic
A secondary de novo detection method was developed to catch strict de novos only lightly checking read depths/AFs etc. This was done as a result of the Hail.de_novo implementation missing at least one clean de novo call in the known results set. 

The use of this naive test has produced a number of noisy de novos we would prefer to have filtered out, and the general performance of the built-in de novo test was fantastic at handling these noisy sites.

Plan: 

1. Remove the twin de novo tests and implement only the Hail version
2. Troubleshoot the known call which couldn't be made with Hail, and if there is a procedural bug find a solution and liase with Hail team to resolve (ref. https://centrepopgen.slack.com/archives/C035QN8C0DN/p1658267361458809)

Possibility this is caused by male X-hets called as homs, and homs are double counted for site AC, which plays a role in the Bayesian prior?

MattWellie:
# Fixes

  - Closes #90 

## Proposed Changes

  - No longer calls the naive de novo logic when generating Category4 - this means there's no need for 4 and 4b
  - Investigation to follow on whether the need for this sub-category in the first place was the result of a fixable bug in Hail's implementation

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/hail_filter_and_label.py
+++ b/reanalysis/hail_filter_and_label.py
@@ -24,7 +24,6 @@ from peddy import Ped
 from cloudpathlib import AnyPath
 from cpg_utils.hail_batch import init_batch, output_path
 
-from reanalysis.homebrewed import custom_de_novo
 from reanalysis.utils import read_json_from_path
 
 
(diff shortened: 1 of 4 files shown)


MattWellie:
the logic for the naive method is retained but not implemented, as a syntax reference for future methods

cassimons (approved):

PR #96, opened 2022-08-19
comp-het bugfix #92

### LINKED ISSUE #86: Remove extra index layer in results output
Current outputs are in the form

```
Sample ID: {
    'key value': {
           Actual variant details
    }
}
```

e.g. 

```
{
    "SAMPLE_1": {
        "7-93105286-T-A__ENSG00000205413__Unsupported": {
            "sample": "SAMPLE_1",
            ...
       }
    }
}
```

This middle key serves no purpose, and is an artefact of the 'cleaning' step, ensuring that the same variant passing > 1 MOI test doesn't lead to duplication in the report. 

This cleaning step should be re-thought, so the final output is simpler to parse (which will help report generation):

```
Sample: [
      Variant1,
      Varaint2,
      ...
],
Sample2: ...
```

### LINKED ISSUE #92: Why didn't these end up in the report?
https://seqr.populationgenomics.org.au/project/R0011_acute_care/saved_variants/family/F000181_fam000364

MattWellie:
# Fixes

  - closes #92
  - closes #86 

## Proposed Changes

  - When the Comp-het analysis is run, the check was not permitting a second-hit variant to be support-only
  - This edit means that a second-hit can now be support-only in compound-het cases
  - Big oversight from me... biallelic hasn't really been treated with the correct leniency up to this point
  - middle-index in JSON output was already removed by an earlier commit
  - this changes the corresponding test fixtures and comparison process to adapt to that

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/comparison/README.md
+++ b/comparison/README.md
@@ -60,12 +60,12 @@ results = {
 }
 ```
 
-In addition, we require a number of other input files from the AIP run being analysed:
+The majority of the input files (AIP outputs, AIP labelled VCF, PanelApp data, config file, pedigree) are located
+relative to the root run folder, so running via the wrapper is simple. The 3 arguments when using the wrapper are:
 
-- PanelApp Data
-- VCF produced by the labelling process
-- MT used to derive labelled variants from
-- Configuration used
+- Seqr Tags CSV
+- MT used as input for AIP run
+- AIP analysis folder
 
 ---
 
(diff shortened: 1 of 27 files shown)


cassimons (approved):

MattWellie:
These changes were already added to main as part of a branch rebased on top of this change. Closing

PR #97, opened 2022-08-22
Comparison update

### LINKED ISSUE #71: Better comparison reporting
Currently the comparison script mainly writes results into runtime logging, and the only thing written to a file is the per-category checking results

Update this so that prior to any hail analysis, there is a top level report (JSON? Whatever) of:

- number of variants from Seqr flagged
- number of variants from AIP
- total matches
- per-sample matches
- per-sample discrepancies
- number of variants missing from final AIP but in the VCF

Then go off and do the hail checking, and write that to a file too

MattWellie:
# Fixes

  - closes #71

## Proposed Changes

  - results are saved into output files, rather than only printing to the Hail logs
  - requires fewer defined inputs - works input paths out by using the run folder root location
  - produces comparison stats for the # of matched and missed Seqr Training Flags

## Example

[This Hail Run](https://batch.hail.populationgenomics.org.au/batches/140982/jobs/1)

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/comparison/README.md
+++ b/comparison/README.md
@@ -60,12 +60,12 @@ results = {
 }
 ```
 
-In addition, we require a number of other input files from the AIP run being analysed:
+The majority of the input files (AIP outputs, AIP labelled VCF, PanelApp data, config file, pedigree) are located
+relative to the root run folder, so running via the wrapper is simple. The 3 arguments when using the wrapper are:
 
-- PanelApp Data
-- VCF produced by the labelling process
-- MT used to derive labelled variants from
-- Configuration used
+- Seqr Tags CSV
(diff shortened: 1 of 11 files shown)


cassimons:
Is this functionality documented in a README somewhere?

cassimons (approved):
LGTM. What would you think about adding a "summary_stats" field to the main output json containing the counts of matched/unmatched for each category?  I know this can be derived from the file easily, but it would make it easy to pull the numbers out when needed?

MattWellie:
@cassimons I've added a straight-up numerical count to the top-line output JSON, and made some changes to the README

PR #98, opened 2022-08-23
Field audit + tweaks

### LINKED ISSUE #58: Conditional acceptance of skip_annotation
The `--skip_annotation` flag allows the annotation phase to be skipped for MTs where we know that the annotation is already correct (keys, values, types, heirarchy).

This isn't checked, so currently we're just trusting that all the correct fields are supplied. 

- Can the matrix_table.describe() results be parsed 
- Can that parsed data be checked against an expected set of annotations

The skip-annotation argument is only really intended for CPG use, as the annotation code is near-identical (and will be shared in future). Worst case scenario for the skip_annotation is that the tool will try and use the data and fail. It would be good to run fail-safe, where missing fields are flagged, and reannotation is requested

MattWellie:
# Fixes

  - closes #58

## Proposed Changes

This PR contains a collection of fixes across a couple of different categories
  - In lots of places there is a non-functional change of `matrix` -> `mt`, to harmonise with the canonical use of mt as a variable name in the hail code base
  - The step MT -> VCF prior to annotation takes AGES, and most of the variants will be filtered out downstream. To reduce overall time and expense for the end2end workflow, sites are removed from the MT where a QC failure filter is shown, and sites are removed if no present samples have alt alleles (possible hangover when extracting the cohort group from a X-cohort joint call). This can be altered to behave differently if required, but for AIP-specific use the filter failures are removed anyway; not doing this removal up front just creates bigger VCF files and longer running annotation jobs.
  - During the annotation phase, the GT stats fields AC/AN/AF are added if they were missing in the VCF (encountered during 1000G processing) (@vladsavelyev could you please check this? In the 1000G dataset example the fields were not populated into the VCF already, so the annotation stage failed [here](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/reanalysis/annotation.py#L105-L107) - I would expect the same point of failure if a non-GATK/Hail generated VCF was supplied as input, and we'd like to be flexible enough to handle non-CPG data)
  - Prior to running the category labelling logic in Hail, an 'audit' is carried out, ensuring that all required fields are present in the correct part of the hierarchy, and are of the correct data type

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/annotation.py
+++ b/reanalysis/annotation.py
@@ -14,6 +14,21 @@ import hail as hl
 logger = logging.getLogger(__file__)
 
 
+def add_call_stats(mt: hl.MatrixTable) -> hl.MatrixTable:
+    """
+    If the GT stats were missing, add now
+    potentially less powerful than running on X-cohort joint-call
+    """
+    mt = mt.annotate_rows(gt_stats=hl.agg.call_stats(mt.GT, mt.alleles))
+    return mt.annotate_rows(
+        info=mt.info.annotate(
+            AC=mt.gt_stats.AC,
+            AF=mt.gt_stats.AF,
+            AN=mt.gt_stats.AN,
+        )
+    )
+
+
 def apply_annotations(
     vcf_path: str,
     vep_ht_path: str,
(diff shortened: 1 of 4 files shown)


cassimons (approved):
Looks great to me.

vladsavelyev (approved):
> During the annotation phase, the GT stats fields AC/AN/AF are added if they were missing in the VCF (encountered during 1000G processing) (@vladsavelyev could you please check this? In the 1000G dataset example the fields were not populated into the VCF already, so the annotation stage failed here - I would expect the same point of failure if a non-GATK/Hail generated VCF was supplied as input, and we'd like to be flexible enough to handle non-CPG data)

Because GATK GenotypeGvcfs adds AN/AC/AF into the joint-called VCF, it's safe to assume the fields are present in the Seqr Loader annotation step. But as you said in your case, AIP should be robust enough to work with any kind of input VCF, so it makes sense to (re-)add those :)

PR #99, opened 2022-08-23
Adjustments for cpg_utils TOML configuration

vladsavelyev:
# Fixes

* Remove old `hail_billing_project` and `hail_bucket` parameters to `query_function` (now read from CPG_CONFIG_PATH)
* CPG_DRIVER_IMAGE is not used and delegated to the toml config as well
* Annotation: skip adding rg37_locus (not doing anything, doing it properly is not supported by batch backend)

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/interpretation_runner.py
+++ b/reanalysis/interpretation_runner.py
@@ -36,6 +36,7 @@ from cpg_utils.hail_batch import (
     output_path,
     query_command,
     remote_tmpdir,
+    image_path,
 )
 
 import annotation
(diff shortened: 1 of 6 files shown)


MattWellie:
@vladsavelyev looks good to me, should the changes be coupled with a cpg-utils dependency update to 4.5.0?

reanalysis/interpretation_runner.py
```diff
@@ -133,7 +134,7 @@ def annotate_vcf(
     batch: hb.Batch,
     vep_temp: str,
     vep_out: str,
-    seq_type: SequencingType | None = SequencingType.GENOME,
+    seq_type: SequencingType = SequencingType.GENOME,
```
illusional, line 137: Now that we have more exome data, I think we should be careful defaulting to Genome and always request it.

MattWellie: Once we start running on Exomes there will be other changes required (at the moment there's no way to supply exome or genome as an argument/config), so I'm happy to assume this now and make an issue for 'things required for exomes'

MattWellie (approved):

PR #100, opened 2022-08-24
remove config toml file

MattWellie:
# Fixes

  - N/A

## Proposed Changes

  - The TOML config file contents aren't important, so just use the template from cpg_utils


test/conftest.py
```diff
@@ -31,7 +32,10 @@
 LOOKUP_PED = os.path.join(INPUT, 'mock_sm_lookup.json')
 
 # can't use a session-scoped monkeypatch, so just set
-os.putenv('CPG_CONFIG_PATH', os.path.join(INPUT, 'cpg_runtime_config.toml'))
+os.putenv(
```
vladsavelyev, line 35: `config-template.toml` loaded automatically in cpg-utils, so you don't need to set it explicitly!

vladsavelyev (approved):
Approving, we can figure out how to run test_apply_annotations on CI separately :)

lgruen:
💯

PR #103, opened 2022-08-29
VEP: slight change in setting up mount

vladsavelyev:
# Fixes

The VEP data location on the `cpg-reference` bucket is parameterised by version now, so needed a slight adjustment to remove the hardcoded `/vep/GRCh38` path.


MattWellie:
@vladsavelyev  `j.image(image_path('vep'))` is still correct with the recent changes?

vladsavelyev:
@MattWellie, yep!

MattWellie (approved):

PR #106, opened 2022-08-31
Update report meta

### LINKED ISSUE #105: Improve quality of report content
Improve the report content so that they 'stand alone' better

- Date the analysis was run 
  - Current this is implicit in the path to the file, but once removed from the GCP bucket this isn't stored
- Extra Panel(s) used, if any - numerical ID, name, version
- Families (count and structure)
  - Table counting the number of Singletons, duos, trios, etc. 
  - Does this have to be using family structure checks? e.g. is 3 are we happy knowing that a family was 3 people, or do we need to strictly check it was a trio? Probably only relevant for sibling duos vs. child-parent duos
- Name of the Cohort/Project being processed
- Path to the MT used during the analysis

MattWellie:
# Fixes

  - closes #105

## Non-issue changes

A couple of QoL changes to helper scripts, useful when setting up new cohorts to run through AIP

- copies all the 'boilerplate' files from the acute-care bucket to the target bucket (CSQ string header line, VQSR header line, pre-panelapp mendeliome gene list)
- when parsing the pedigree and internal:external ID mapping from metamist, includes a `--copy` argument to additionally push those files directly from local to GCP (saves 2 manual copying operations, will fail if local user doesn't have write permission/is not auth'd)

## Proposed Changes

  - does a better job of logging the names and versions of all panels parsed as part of the analysis (name/version/id) in the panelapp output
  - counts all family structures common between the VCF and pedigree, and writes into the results metadata
  - counts male/female/affected and writes into HTML as a table
  - stores top-level metadata in the output config JSON which is copied into the run output folder
  - use the 'this run' config when running downstream stages
  - adds a trio of tables to the output HTML:
  
1. Panels: The name, ID, and version number of all panels used in the analysis
2. Meta: cohort, run time, and input file used
3. Families: count of number of families of each size, male/female, affected participants 

## Further improvements

1. instead of `don't run step X if the output exists`, find a better way - e.g. incorporate the hash of all extra panel IDs in the panelapp data, or always run a panelapp query and hash the content to decide whether to re-run the hail phase (e.g. check if green genes have changed) - this would require running the panelapp task before the batch is set off (currently they are both dependent jobs, and the hail stage can't be skipped in response to the panelapp result)
2. incorporate the cohort name into the html/json output name? 

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/helpers/cohort_prep.sh
+++ b/helpers/cohort_prep.sh
@@ -13,5 +13,6 @@ mkdir -p "inputs/$COHORT"
 python helpers/pedigree_from_sample_metadata.py \
     --project "${COHORT}" \
     --output "inputs/${COHORT}/${TODAY}_${SUBSAMPLE}_percent" \
-    --plink --hash_threshold \
-    "${SUBSAMPLE}"
+    --plink \
+    --hash_threshold "${SUBSAMPLE}" \
+    --copy
--- a/helpers/file_prep.sh
+++ b/helpers/file_prep.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -x
+
+# 'canonical' files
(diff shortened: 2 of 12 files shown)


cassimons (approved):
LGTM

PR #107, opened 2022-08-31
Non-functional refactoring of VEP

vladsavelyev:
## Proposed Changes

- [x] Use new `cpg_utils.workflows` module for batch resources and other utilities
- [x] Use `cpg_utils.Path` instead of `CloudPath` (it would make it work for local paths)
- [x] Update `split_intervals`, use the source intervals lists from config
- [ ] Move `split_intervals` and `gather_vcfs` into `cpg_utils`?
- [x] Sync with Seqr Loader to potentially make the former import AIP
- [ ] One issue in Seqr Loader importing AIP is that AIP strictly pins packages, which means that packages would conflict. @MattWellie, what do you think of dropping pins and releasing AIP as a pypi package?

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


MattWellie:
@vladsavelyev following last week's discussion, is this still a live PR, or are we going to move to a dependency directly on `production-pipelines`?

MattWellie (approved):

MattWellie (approved):

MattWellie:
@vladsavelyev I'm going to merge this, but eventually I think this code will be imported from production-pipelines

vladsavelyev:
Oh sorry @MattWellie, I missed your question question 6 days ago... 
I'm going to look into importing this from prod-pipes!

PR #108, opened 2022-09-01
Panel hpo

MattWellie:
# Fixes

  - contributes to #109 

## Proposed Changes

These functions are all buried in a single file:

1. query PanelApp for panels with linked HPO terms:
  - parses panelapp 'overview' endpoint, using the `relevant_disorders` values
  - indexes all panel IDs by the HPO terms they relate to (only keep panels with HPO data)
  - provides this as a dictionary

2. HBO term parser:
  - obonet/networkx to parse the jax HPO .obo file
  - takes an obo file, and parses it into a graph object

3. querying metamist for samples and corresponding HPO terms
  - Ok, so this isn't currently working... because the Metamist API client isn't working for the seqr endpoint
  - instead this is working based on a dump file (go to the swagger page, query for a cohort, download the JSON, use that as input)
  - returns a dictionary indexed on Participant ID, including both FamilyID and a list of all HPO terms (or an empty list)

4. matchy matchy process between the HPO terms present in a cohort, the panelapp panels with HPO terms
  - takes all unique HPO terms found in the cohort, uses a recursive/tree-traverse to do a fuzzy match and finds any relevant panels within `n` layers of HPO heirarchy
    - accounts for the deprecated HPO terms, re-shuffling to search for the replacement terms
    - accounts for the multiple-parent terms to search all parents

5. re-match to take all the panels-per-HPO term identified in the cohort and add them to all relevant individuals
  - takes all HPO terms and their matching panels (or empty set)
  - for each participant, iterate over all HPO terms recovered from metamist
    - if the HPO term was associated with any panels, add to that participant's personal collection
  - dump the dictionary to a file, re-writing the sets as lists for the JSON formatter to work

End result of 1-5: a single dictionary of

```python
{
   'participant_id': {
        'family_id': str,
        'hpo_terms': ['list', 'of', 'terms'],
        'panels': ['list', 'of', 'panel_id']
    },
    ...
}
```

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

## TODO

- needs documentation
- needs integration with the panelapp query and the result filtering
- should accept alternative participant-hpo sources


MattWellie:
Up for discussion whether at each stage we should check for a node(HPO term)'s parent, then also find all their immediate children. This would be a fuzzier search, may be worth playing with this once more terms are populated and seeing how the results feel

MattWellie:
@cassimons this is a first stopping point for the HPO~Panel~Participant matching

Following PRs will separately deal with changes to how panels are pulled from PanelApp, and how the AIP results are post-filtered based on those individual gene lists

cassimons (approved):
Looks great.

helpers/hpo_panel_matching.py
```diff
@@ -0,0 +1,286 @@
...
+
+MAX_DEPTH: int = 3
+PANELAPP_SERVER = 'https://sample-metadata.populationgenomics.org.au'
+TEMPLATE = (
+    f'{PANELAPP_SERVER}/api/v1/participant/{{dataset}}/individual-metadata-seqr/json'
+)
+
+
+HPO_RE = re.compile(r'HP:[0-9]+')
+OBO_FILE = os.path.join(os.path.dirname(__file__), 'hpo_terms.obo')
```
MattWellie, line 35: I think this should be altered - currently the process is 100% reliant on an HPO.obo file being committed into the repo, which is ~8MB. Instead re-write this so that the path to the file is an argument

PR #110, opened 2022-09-14
Additional findings

MattWellie:
# Fixes

  - N/A

## Proposed Changes

  - This is a bit of a sidebar, first draft of a parser and hail-labeller for additional findings
  - Parser for the additional findings variant CSV (which will find `All Variants`, `Specific Consequences` and `Specific Protein Changes`) 
  - AIP parser produces a format which will be interchangeable with the PanelApp parser for downstream MOI checking
  - Hail labelling script shares as many methods as possible with the original hail script (created a separate `hail methods` script to import common content from)
  - Hail stage produces a VCF with the same label syntax as AIP-main, so we can pass the VCF straight through to the AIP results checking phase

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


MattWellie:
I've just realised how heinous this PR is... it incorporates a number of changes Vlad has made in the cpg_utils package so there is no usage of `AnyPath` any more, which occurs all over the codebase

cassimons:
Can we chat about this? I am quite keen to separate this into its own repo to keep the boundaries clean as it is not part of the AIP project.

MattWellie:
@cassimons happy to discuss - separate or not it can use so much of the AIP codebase that it makes sense to share. @vladsavelyev  was suggesting that AIP be pushed as a pypi package (for annotation reasons); if I sort that out, sharing code can be done without explicitly being in the same repository, that might be a good solution

Mostly I just wanted to hammer out some logic around digesting the ACMG data format & passing it through hail filters such that the AIP MOI-tests can be used transparently for additional findings.

cassimons:
I agree it would be great to share as a good chunk of the code. I am just keen to not cross the streams of the different sets of stakeholders for the two projects.  If we can go with seperate projects but importing the core logic from the AIP then that would be perfect.

PR #117, opened 2022-09-26
add incidentalome query and tests

### LINKED ISSUE #116: Integrate the Incidentalome into new panel queries
Related to #109 

Add a way to integrate the Incidentalome data into the panelapp query

Incidentalome is a special case where we want to remove a subset of the genes based on CPG-specific rules (for now, this may change in future).

MattWellie:
# Fixes

  - Closes #116 

## Proposed Changes

  - adds a non-default flag option to query for the panelapp stage
  - if used, this flag means that the core panel data becomes Mendeliome + Cardiac Incidentalome instead of just the Mendeliome
  - 'cardiac' Incidentalome is determined using flags
  - The Flag-matching mechanic is available for any other potential flags

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/design_docs/PanelApp_interaction.md
+++ b/design_docs/PanelApp_interaction.md
@@ -17,6 +17,12 @@ During an analysis run, the [Query Script](../reanalysis/query_panelapp.py) is u
 for the Mendeliome panel (gene ENSG, symbol, and Mode of Inheritance). This is saved as a dictionary indexed on
 GRCh38/Ensembl 90 ENSG.
 
+## Incidentalome Flag
+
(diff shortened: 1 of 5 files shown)


MattWellie:
Following discussion with @cassimons, this specific addition is cancelled

The Incidentalome is used as a complement to the Mendeliome, containing genes which could cause unintended additional findings. 

As this set of PRs is allowing for the use of phenotype-specific panels, the Incidentalome is no longer relevant.

```
i.e. instead of 'taking the cardiac genes from the incidentalome and applying to all participants', which this PR adds in, we are instead aiming for 'if this participant has a cardiac phenotype, add all cardiac-relevant genes from the appropriate panel'
```

This negates the need for the incidentalome entirely.

Issue #113, opened 2022-09-21
Remove some config file arguments


MattWellie:
# Issue

There are 2 files which are currently required in a bucket for AIP to run end-to-end:

1. VQSR Filter header line, required for converting a MT into a VCF that can be parsed without error
2. CSQ String, required when writing the output VCF from the Hail filtering stage

Both of these files complicate deployment/triggering analysis in new buckets/datasets, as they have to be copied into the namespaced environment for the data, and are trivial one-line files.

# Proposal

Alter the flow to write the required file out at runtime (e.g. into temp), then use it from that path

Hard-code the contents of the CSQ line into the CSQ-attribute generating method - This shouldn't be present in config, as changes to the method should lead to changes in the line content. 

Keep the VQSR line in config? 

- If null, do nothing (fine when a dataset was not processed using VQSR)
- If present, write to a file and then add in the VCF header

PR #118, opened 2022-09-28
adds fixes from 113 independent of panel feature

MattWellie:
# Fixes

  - issues stemming from #113 

## Proposed Changes

  - fixes the earlier header file commit 
  - this is also fixed in the #108 #111 #112 #115 PR cascade, but this is being fast-tracked separately so that the MSFT deployment isn't broken


cassimons (approved):
LGTM

Issue #128, opened 2022-10-10
Shift config from JSON to Toml


MattWellie:
Currently config uses a manually passed JSON file, which adds a lot of coupling in method calls.

Instead of that, use a TOML file, and use the CPG-utils `config` functionality to absorb the config content into the broader environment config.

Reference: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1665451208949389

Increments:

1. Split current whole-run config examples into 2 sub-configs
  - AIP-general purpose (thresholds, filter settings, report colours)
  - Cohort-specific (e.g. sample lookups, seqr project...)
2. convert JSON configuration files to TOML
3. update analysis-runner call to pass the general and cohort-specific configs as arguments
4. change ALL config usage to reference the global config using `get_config`

Note: This breaks up the config changes into 4 separate steps, but step 4 is obviously 99% of the effort and 100% of the code changes... I can't see an easy way to mitigate that, unless both config files are supplied and different parts of the codebase use different approaches, which seems silly

MattWellie:
Update: Following a re-run of the `ag-hidden` cohort, some variants have changed in the final report

[PREVIOUS](https://main-web.populationgenomics.org.au/ag-hidden/reanalysis/2022-10-06/summary_output.html)

[LATEST](https://main-web.populationgenomics.org.au/ag-hidden/reanalysis/2022-10-12/summary_output.html)

Gained:
A2234017 Cat. 2 FRMD5 var.

Lost:
A1534021 Cat. 1 GALT var.


There are different panel versions, but that probably won't explain this difference

MattWellie:
Ok... Maybe it's absolutely fine. The older version of the Mendeliome (by 6 days) has no FRMD5 gene entry, and GALT is Monoallelic. 
In the latest version FRMD5 is present, and GALT is biallelic. This explains both discrepancies

PR #129, opened 2022-10-11
Config to TOML

MattWellie:
# Fixes

  - Starts #128

# Description

This is the base PR in a chunky stack of PRs - once every following PR is collapsed into this one, the following behaviours should be changed:

1. Config is not provided as cloud-located JSON files; instead the configuration is split into two sub-components - generic and cohort-specific. These are read locally, and the contents are sent to the analysis-runner server through the miracle of `cpg_utils.config`
2. Instead of passing a configuration file to each different stage of the pipeline, the run-specific meshed config is read using the `get_config()` operations (meaning far fewer arguments have to be specifically passed between method calls)
3. Tests will be updated to use indirect configuration access, via session-level test fixtures
4. The comparison process needs updating to reflect the configuration file changes

## Proposed Changes

  - breaks the core config into two separate files - generic, and cohort specific
  - includes a TOML version of each file
  - this is the first of many

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


design_docs/Hail_Filter_and_Label.md
```diff
@@ -66,7 +66,7 @@ Annotated Joint-called MatrixTable. Annotations applied either by, or consistent
 
 ### Process
 
-1. Read reference files (PanelApp data, [Configuration](../reanalysis/reanalysis_conf.json))
+1. Read reference files (PanelApp data, [Configuration](../reanalysis/reanalysis_global.json))
```
vladsavelyev, line 69:
```suggestion
1. Read reference files (PanelApp data, [Configuration](../reanalysis/reanalysis_global.toml))
```

vladsavelyev (approved):

PR #136, opened 2022-10-13
Update run_reanalysis.sh

MattWellie:
Changes to runner template, reflects the latest CLI changes

changes to arguments/flags to reflect latest CLI args

### DIFF
--- a/reanalysis/run_reanalysis.sh
+++ b/reanalysis/run_reanalysis.sh
@@ -13,6 +13,6 @@ analysis-runner \
   -o "reanalysis/${DATE}" \
   --access-level test \
   reanalysis/interpretation_runner.py \
-    --input_path gs://cpg-acute-care-test/reanalysis/2011-11-11/prior_to_annotation.vcf.bgz \
-    --panel_genes gs://cpg-acute-care-test/reanalysis/pre_panelapp_mendeliome.json \
-    --plink_file gs://cpg-acute-care-test/reanalysis/acute-care-plink.fam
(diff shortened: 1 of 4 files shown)


reanalysis/run_reanalysis.sh
```diff
@@ -13,6 +13,6 @@ analysis-runner \
   -o "reanalysis/${DATE}" \
   --access-level test \
   reanalysis/interpretation_runner.py \
-    --input_path gs://cpg-acute-care-test/reanalysis/2011-11-11/prior_to_annotation.vcf.bgz \
-    --panel_genes gs://cpg-acute-care-test/reanalysis/pre_panelapp_mendeliome.json \
-    --plink_file gs://cpg-acute-care-test/reanalysis/acute-care-plink.fam
+    --i gs://cpg-acute-care-test/reanalysis/2011-11-11/prior_to_annotation.vcf.bgz \
+    --pedigree gs://cpg-acute-care-test/reanalysis/acute-care-plink.fam \
```
vladsavelyev, line 17: `.fam` and `.ped` are interchangeable formats, right?

vladsavelyev (approved):

reanalysis/run_reanalysis.sh
MattWellie: Yes, they should be. This was from some dodgy formatting early on. I was allowing empty strings for missing parents in the pedigree file. This works for some parsers, but broke Hail. I moved to the `.fam` naming convention to indicate that this required a PLINK-fomatted PED file.

The whole thing is dumb, so I've changed the PED file builder. So long as the file is correctly formatted, `.ped` or `.fam` should be fine

PR #138, opened 2022-10-14
Write human results in analysis

### LINKED ISSUE #137: Move default write path to `-analysis` bucket
The output paths for AIP are currently in the `main` data bucket. This bucket is not accessible to human users, so for ease of access and future debugging, move this output path to `main-analysis`, in keeping with the [CPG policy](https://github.com/populationgenomics/team-docs/tree/main/storage_policies#analysis-gscpg-dataset-maintest-analysis)

MattWellie:
# Fixes

  - closes #137 

## Proposed Changes

  - Writes human-useful files to the `-analysis` subset bucket instead of directly in main
  - This fragments the output across more folders, but makes it user-accessible as/when required, e.g. debugging
  - Summary JSON, Pedigree(s), and PanelApp data
  - Takes the analysis suffix from config, where it can be missing/None (i.e. if the deployment doesn't have a separate analysis bucket, results will be written to the regular location)

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/reanalysis_global.toml
+++ b/reanalysis/reanalysis_global.toml
@@ -1,6 +1,7 @@
 [buckets]
 web_suffix = "web"
 tmp_suffix = "tmp"
+analysis_suffix = "analysis"
 
 [moi_tests]
 description = "thresholds for different filters during the MOI checks"
--- a/reanalysis/interpretation_runner.py
+++ b/reanalysis/interpretation_runner.py
@@ -500,8 +500,10 @@ def main(
 
     # write pedigree content to the output folder
     to_path(pedigree).copy(
-        output_path(f'pedigree_{EXECUTION_TIME}.fam'),
-        get_config()['buckets'].get('analysis_suffix'),
+        output_path(
+            f'pedigree_{EXECUTION_TIME}.fam',
+            get_config()['buckets'].get('analysis_suffix'),
+        )
     )
     # endregion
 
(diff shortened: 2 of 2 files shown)


reanalysis/reanalysis_global.toml
```diff
@@ -1,6 +1,7 @@
 [buckets]
 web_suffix = "web"
 tmp_suffix = "tmp"
+analysis_suffix = "analysis"
```
lgruen, line 4: Do we really need to make this configurable? Particularly if we can resolve https://github.com/populationgenomics/automated-interpretation-pipeline/issues/122#issuecomment-1276894986 more generally?

lgruen (approved):

reanalysis/reanalysis_global.toml
MattWellie: Possibly not, but it permits a user to leave out that index completely if we want `suffix=None` to be passed to `output_path`, depending on infrastructure setup.

I definitely think it can be improved upon, e.g. overriding the bucket names using @illusional 's format string suggestion

vladsavelyev: Sorry, only catching up with this discussion. 

I like Michael's suggested toml section. But noting that there is some existing code for cloud paths in cpg_utils, which we can extend upon, or reuse the existing terminology. Happy to look into that more next week

PR #146, opened 2022-11-04
Remove image path

MattWellie:
# Fixes

  - Driver image is no longer present within Images in config, so we need to find it from the `workflow` group

## Proposed Changes

  - remove the `image_path` call
  - replace with grabbing the image path from `[workflow]`


vladsavelyev (approved):

vladsavelyev:
> Driver image is no longer present within Images in config, so we need to find it from the workflow group

You can still add an `[images]` section into the AIP-specific config with any arbitrary images, and `image_path` will recognise it

MattWellie:
> `[images]` section into the AIP-specific config 

💯 ! For this usage I just want to make sure I'm targeting the latest driver image for now, but good to remember for custom tool images

PR #151, opened 2022-11-18
Clinvar but better

### LINKED ISSUE #147: Apply manual Clinvar curation
Review all clinvar raw data, and implement custom per-variant resolution. The intention is to have these per-variant results available to replace the default clinvar annotations provided by the Broad, and run AIP with those alternate decisions.

For reference: https://miro.com/app/board/uXjVPcOGkd8=/

These are some results we consider accurately resolved on the assc. branch:

- https://ncbi.nlm.nih.gov/clinvar/variation/210109/ : `Uncertain` rating was prior to mainstream ACMG usage, so only the `Benign` rating is relevant, rather than being rated as conflicting
- https://ncbi.nlm.nih.gov/clinvar/variation/210112/ : `Uncertain` (1), `Likely_benign` (3), `Benign` (1), we're happy to resolve this as `Benign` instead of conflicting
- https://ncbi.nlm.nih.gov/clinvar/variation/245759/ : `Uncertain` rating was prior to mainstream ACMG usage, so only the `Pathogenic` rating is relevant, rather than being rated as conflicting
- https://ncbi.nlm.nih.gov/clinvar/variation/10/ : `Pathogenic` (24), `Uncertain significance` (2), `Benign` (1), `Likely benign` (1) - we're happy to have this resolve to `Pathogenic`

1. Create a rules flow to express the new decisions we want to make at an individual allele (https://miro.com/app/board/uXjVPcOGkd8=/)
2. Run this new variant summary to generate a table of our variant decisions
3. Incorporate this new variant summary into the joint-call through annotation
4. Re-run AIP using these new annotations

## Notes

- these private annotations should be optional, defaulting to the VEP-supplied versions if the extra table isn't referenced

### LINKED ISSUE #149: Private Clinvar
See https://github.com/populationgenomics/automated-interpretation-pipeline/issues/147

Summary: We are not always happy with the top-line clinvar summary decisions, and would like to make our own from the raw collection of submissions 

1. Create a rules flow to express the new decisions we want to make at an individual allele (https://miro.com/app/board/uXjVPcOGkd8=/)
2. Run this new variant summary to generate a table of our variant decisions
3. Incorporate this new variant summary into the joint-call through annotation
4. Re-run AIP using these new annotations

## Notes

- these private annotations should be optional, defaulting to the VEP-supplied versions if the extra table isn't referenced

MattWellie:
# Fixes

  - Closes #147 
  - Closes #149 

## Context

- [Acute-Care results with the MOI changes but unchanged Clinvar](https://test-web.populationgenomics.org.au/acute-care/reanalysis/2022-11-11/summary_output.html)
- [Acute-Care results with MOI changes AND clinvar changes](https://test-web.populationgenomics.org.au/acute-care/reanalysis/no_private/summary_output.html)
- [Previous results, same joint call, no changes](https://test-web.populationgenomics.org.au/acute-care/reanalysis/2022-09-23/summary_output.html)

## Proposed Changes

This is PR is a superset of #148 - #148 is just the single script that generates a manual summary file from the raw submissions. That PR will be closed, but could be cleaner when viewed in isolation 🤷 I meant to separate this PR into stacked stages but forgot. From the other PR: 

  -  This script takes pointers to a file of individual submissions and the file of `variant <-> locus <-> alleles`
  - Clusters all submissions by allele, applies the manual reconciliation process [here](https://miro.com/app/board/uXjVPcOGkd8=/) to all submissions on a single allele
  - Makes a decision based on retained submissions at an allele, discarding alleles which don't cleanly fall into Benign or Pathogenic conclusions (See `Issues`)
  - Sorts all decisions based on a dual-layered sort (contig, position)
  - Iterates over contigs in order, and generates a per-contig Hail Table from the list of dictionaries
  - Progressive merging of the single-contig hail tables into one large table
  - Write that out to GCP

Additional Content:
  - The Hail field-audit content has now been pulled out into a separate file, making it more accessible to other usages and reducing the overall line count for the Hail methods file.
  - Removed the AB ratio test and a string formatting method, neither were in use
  - Hail/Classification stage now understands a `private_clinvar` optional config entry. If this exists, the Hail table it points to is loaded, and the filtering proceeds with the imported annotations. If it doesn't exist, the extant annotations from VEP are used (as before)
  - Filtering is done to remove any entries where the clinvar status is Benign and stars > 0
  - 2 boolean Summary values are created: `clinvar_aip` (Pathogenic at any confidence) and `clinvar_aip_strong` (Pathogenic and > 0 stars) - these are checked in downstream logic instead of the triple condition `clinvar_significance.lower() contains pathogenic and doesn't contain conflicting and stars > 0` - hopefully this should speed up the category checks, and slightly offset the longer duration from the new clinvar lookup
  - When filtering against gnomad rare, we now allow variants over the threshold if they have the `clinvar_aip` flag

MOI filtering changes:
  - The Gnomad Rare/Hom count/Hemi count conditions are applied to both candidate variants in a compound-het, instead of only the `primary` variant
  - For every MOI-Gnomad AF test, we allow Cat 1 to qualify even if high af

TESTS

- lots of tests added and altered to incorporate the above changes

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

## Future

- A lot of the Clinvar loading/filtering could be optimised using Pandas - might be a good side project for some pandas-savvy individual with time to burn 🐼 
- Gnomad v2 AF checks aren't currently included. We are checking against both gnomad exomes & genomes for Hom/Hemi counts. Within hail we could take the highest AF value in Gnomad 2/3(/4) and use this for downstream AF checks if we definitely want to consider the worst AF across multiple datasets.


clinvar/conflict_huntr.py
```diff
@@ -0,0 +1,545 @@
...
+        filename = tempfile
+
+    with gzip.open(filename, 'rt') as handle:
+        for line in handle:
+            if line.startswith('#'):
+                continue
+            yield line.rstrip().split('\t')
+
+
+def county_county(subs: list[Submission]) -> Consequence:
```
cassimons, line 206: Um, is this a thing?

cassimons (approved):

clinvar/conflict_huntr.py
MattWellie: Hah! good spot. I changed this to a better name in the parallel PR, but didn't bring it over

MattWellie:
I'm closing this... but there is a weird semi-non deterministic unit test issue I couldn't really understand

- There is a session-scoped test fixture which reads a 2-variant VCF, annotates in some blank clinvar annotations, and writes the resultant MT to a temp location -> the fixture returns this temp path
- 2 tests use this fixture, one which tests the default behaviour (no private clinvar annotations added), and one which makes clinvar annotation tables on the fly and uses them as annotation
- Even though each of these test runs reads the MT from disc fresh, the annotations persist from the custom annotation case into the default case UNLESS the default case is first in execution order.

I suspect this is due to hail's runtime being scoped to the full pytest execution rather than each test, and there is crossover from the referenced objects.

Issue #152, opened 2022-11-18
Use Seqr to provide data directly


MattWellie:
# Metadata lookups? WE'LL DO IT LIVE

![Image](https://user-images.githubusercontent.com/5886855/202821675-7de4aa1c-dc94-421f-8f07-f46ca98db42e.png)

@illusional has provided an API view onto the metadata content, accessible through a couple of endpoints. One endpoint in particular only exists on `seqr-reanalysis-dev` at the moment @ 

`https://<SEQR_INSTANCE>/api/project/<PROJECT>/samples/sa`

e.g.

`https://seqr-reanalysis-dev.populationgenomics.org.au/api/project/R0003_acute_care_testing/samples/sa`

Directly accessing the data from this endpoint could be used to replace the current cache'ing of json to link between CPG IDs and seqr individual/family IDs. Put together a couple of usage examples to demonstrate how this could be useful

## Issues

These endpoints are restricted to specific service accounts, so will require a specific service user authentication... Is that true? Ideally we'll be able to run this as the same user we authenticate as through analysis runner, rather than doing some weird things with authentication at runtime to swap users

Writing code that uses this will make some invalid assumptions - every Seqr instance will have these novel endpoints.

## MVP

For demo purposes, write a single script (which could be swapped in as an AIP component) to do the following:

1. Start with a project ID
2. Query Metamist for the pedigree details
3. Query Metamist for the corresponding project(s) in seqr
4. Phrase Seqr queries (using a second, separate GCP account for now) to obtain the family ID in seqr for each sample
5. Allow for possibility of each sample mapping to multiple sequence IDs, and possibility of presence in multiple seqr projects, each with a different family ID

Output: 

- Pedigree file (in terms of CPG ID)
- Dictionary (toml/json/oxford) of each seqr project for the metamist project, and all the corresponding sample -> family mappings obtained
- Additional dictionary of all CPG -> External ID mappings....

Showerthought - can the external ID just be bonded onto the individual's entry in the PED file as an extra column? This could be ignored or parsed deliberately, instead of taking an extra lookup file - no external ID column, NO SALE

illusional:
We could give the seqr service accounts access. Though we'll have to do it manually, and case by case specific datasets, ie acute care for test or something.

Technically they have access to the same level of genetic data anyway.

@lgruen, one other way we could achieve this is proxy requests through metamist and do the access checks there.

illusional:
[...]

PR #153, opened 2022-11-23
scripted prototype for seqr interactions

MattWellie:
# Fixes

  - relevant to #152 

This is not polished or robust, it's just a proof of concept

## Proposed Changes

  - Dependent on https://github.com/populationgenomics/seqr/pull/167
  - prototype post-aip-run script so that (if the `@service_account_access` auth was live):
  
1. query metamist for the correct seqr project to use, or die if not found
2. query seqr for CPGid -> seqr family ID   lookup
3. query seqr for all current flagged variants in the project
4. digest a set of AIP results
5. find all results in the AIP set which don't already exist as flags
6. push each of those flags as a new saved variant

I've played around with this on a local seqr deployment, and tested most of these methods/api calls against that instance. Without an elastic connection it's not been possible to confirm that the pushed tag work perfectly

## Not done: 

There's some many2many mappings behind the scenes, so the same tag can be shared by 2 families , or a variant can be linked to multiple tags. This is a basic implementation, but there's some slightly more involved syntax that could be used to update/extend existing tags instead of always creating new ones.

Also a fair bit of this logic could be done on the seqr ORM side, but I gave up on that because tracing joins on the backend was giving me a migraine.

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


reanalysis/seqr_interactions.py
```diff
@@ -0,0 +1,379 @@
...
+    # WALRUS
+    if results := len(projects) != 1:
+        raise Exception(f'Expected one project for {project_dataset}, found {results}')
+
+    project = projects[0]
+
+    if proj_seqr_key in project['meta'] and project['meta'].get('is_seqr'):
+        return project['meta'].get(proj_seqr_key)
+
+    return None
```
illusional, line 91: You're already getting seqr projects above:
```suggestion
    return project['meta'].get(proj_seqr_key)
```

reanalysis/seqr_interactions.py
```diff
@@ -0,0 +1,379 @@
...
+
+        # now check to remove redundancy
+        existing = extant_flags[family_id]
+
+        for variant_id, tags in variants.items():
+            for tag in tags - existing.get(variant_id, set()):
+                flags_to_add.append(
+                    {'family': family_id, 'tag': tag, 'variant': variant_id}
+                )
+            continue
```
illusional, line 288: Redundant continue?

illusional (commented):
Looking good!

reanalysis/seqr_interactions.py
MattWellie: Heh, I was wondering if there was any possibility of having a populated project and `is_seqr=False`, but that wouldn't make any sense...

reanalysis/seqr_interactions.py
MattWellie: Yup!

reanalysis/seqr_interactions.py
```diff
@@ -0,0 +1,375 @@
...
+    proj_seqr_key = f'seqr-project-{get_config()["workflow"]["sequencing_type"]}'
+
+    # filter the returned projects
+    projects = [
+        proj
+        for proj in ProjectApi().get_seqr_projects()
+        if proj.get('name') == project_dataset
+    ]
+
+    # WALRUS
```
cassimons, line 82: IMHH the walrus should die in a fire 🔥 🦭 🔥

cassimons (approved):
Very cool!

cassimons:
Can you retract a 'approve? Too much muscle memory...

MattWellie:
Closing for now

Issue #154, opened 2022-11-23
Time-filter PanelApp results


MattWellie:
Simulate time travel by identifying the genes which were present on the mendeliome/phenotype panels at different points in time - use this to simulate the gene list as if we were live-querying at those times. 

- The MOI was added recently, so take the current MOI at all times

e.g. what was the latest mendeliome version at 2021-01-01, 2021-07-01, 2022-01-02, 2022-07-01, and what genes were green at those times

This information should be incorporated alongside the time-restricted clinvar results to produce AIP runs which simulate results at given points in the past

MattWellie:
This is being provisionally marked as done based on https://centrepopgen.slack.com/archives/C035QN8C0DN/p1670194659034299

PR #164, opened 2022-12-01
Filter to new results

### LINKED ISSUE #155: Filter out previously seen results
AIP requires a way of filtering to new results:

- generate a full set of results
- take a set of previous results as context
- filter out any current results which were previously seen
- write out two files
    - cumulative results
    - incremental unique results from this run

Important factors:

- each previously seen variant must contain the categories it was assigned
  - e.g. if a variant was Cat 1 before, and is Cat 1 & 4 now, report should show the variant
  - TBD if the report should show all categories or just new categories

- each run should update the cumulative results
  - within GCP, assuming standard permissions, file overwriting is not allowed
  - config should have a folder, and we find the latest file within that folder
  - latest can be determined using `to_path('PATH').stat().st_mtime`

MattWellie:
# Fixes

  - closes #155 
  - enables #154

## Proposed Changes

@cassimons in particular, can you check that this implementation feels fit for purpose?

  - This is a pretty THICC change, with 2 Cs
  - Introduces the concept of filtering current results by removing previously seen results
  - Admittedly this all needs documenting, but the docs need a user-focused overhaul anyway so this will be done at that time

1. The config can now contain the key `dataset_specific.historic_results` - this will be the path to a cloud folder containing cumulative results files - this would be a candidate use of metamist's `analysis` objects if we wanted to tie AIP to CPG.
2. When AIP runs, the results will be cleaned against these prior results:
    a. The latest file in the `historic_results` folder will be identified using the `stat().st_mtime` attribute of the files (most recent modification time)
    b. This check doesn't fail if the directory is empty, but in that situation `None` will be returned instead of a file path
    c. If no file was found, the current data is saved in a minimal representation (`See Below`) into that folder, and all current results will reach the report
    d. If a file is returned, it is loaded as a dictionary, and the current data is filtered against it (`See Below... again`)
    e. The historic data is updated to include all new results not previously seen, and is saved back into the same folder (under a new file name, which will include the date/time to prevent write clashes)
    f. Filtered results are returned to be made into a report

I could write the results of any given run out before and after this filter is applied, if that might be of interest.

## See Below

The historic data doesn't require all the variant details, so we only save 3 items per variant: coordinates (in chr-pos-ref-alt form), assigned categories, and variants supporting (comp-het). This is the only data written as the 'historic' results, which means manually adding prior external results could be prepared as a simple bit of JSON (probably not a likely use case). Each time a new reportable variant is seen, this data is extended - if we see the same variant but in a new category, the on-file cumulative representation only grows by ~1 Byte.

```python
{
    'sampleID': {
        'chr-pos-ref-alt': {
            'categories': ['1', '4', ...],
            'support_vars': ['chr-pos-ref-alt', ...]
        },
        ...
   }
}
```

## See Below... Again

When we filter current results against historic results, it goes something like this:

1. Sample: Have we seen variants for this sample previously? If not, keep everything
2. Variant: Have we seen `this` variant in this sample previously? If not, keep this variant in full
3. Partners: When we previously saw this variant - was it supported in a comp-het?
    - if yes, but this time it's supported by a different comp-het pair, keep the whole variant and all categories
    - if yes, and the same partner is seen again, go to checking categories
4. Categories: Comparing current results with the previous sighting of this variant:
    - are there any new categories assigned? If so, keep only the new categories
    - if not, we don't want to see this variant again - remove from the report
 

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/interpretation_runner.py
+++ b/reanalysis/interpretation_runner.py
@@ -100,13 +100,14 @@ def set_job_resources(
     )
 
 
-def mt_to_vcf(batch: hb.Batch, input_file: str):
+def mt_to_vcf(batch: hb.Batch, input_file: str) -> hb.batch.job.Job:
     """
     takes a MT and converts to VCF
     :param batch:
     :param input_file:
     :return:
     """
+
     mt_to_vcf_job = batch.new_job(name='Convert MT to VCF')
     set_job_resources(mt_to_vcf_job)
 
@@ -313,7 +314,9 @@ def main(
             ANNOTATED_MT = input_path
 
         else:
-            prior_job = mt_to_vcf(batch=get_batch(), input_file=input_path)
+            if not to_path(INPUT_AS_VCF).exists():
+                prior_job = mt_to_vcf(batch=get_batch(), input_file=input_path)
+
             # overwrite input path with file we just created
             input_path = INPUT_AS_VCF
     # endregion
--- a/reanalysis/interpretation_runner.py
+++ b/reanalysis/interpretation_runner.py
@@ -41,7 +41,7 @@ from cpg_workflows.jobs.vep import add_vep_jobs
 from utils import FileTypes, identify_file_type
 
 # exact time that this run occurred
-EXECUTION_TIME = f'{datetime.now():%Y-%m-%d %H:%M}'
+EXECUTION_TIME = f'{datetime.now():%Y-%m-%d_%H:%M}'
 
 # static paths to write outputs
 INPUT_AS_VCF = output_path('prior_to_annotation.vcf.bgz')
--- a/reanalysis/moi_tests.py
+++ b/reanalysis/moi_tests.py
@@ -339,7 +339,6 @@ class DominantAutosomal(BaseMoi):
                     gene=principal_var.info.get('gene_id'),
                     var_data=principal_var,
                     reasons={self.applied_moi},
-                    supported=False,
                     flags=principal_var.get_sample_flags(sample_id),
                 )
             )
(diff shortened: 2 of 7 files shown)


MattWellie:
I've just run this twice on the validation dataset (a run which had previously completed, so only the MOI/reporting stage ran) 

- first run the file `gs://cpg-validation-test-analysis/reanalysis/historic_results/2022-12-01_04:57` was written, containing the only sample & variant present on the final report
- running the exact same run again, the cumulative results were read, all variants were filtered out, and the report is now empty

cassimons:
I think this logic is sound for alerting. However, an alternative final output I believe we should also support is exporting all variants but adding a field to each that records the date of the report in which they were first identified (for this category?). This means an analyst can a) quickly check for new variants but b) seamlessly progress to look at all previous variants.

I think your existing historic_results format would work fine for this, just by adding the date_reported field/s.

 We could also think about adding this date-based recording to our gold stars - not exactly sure how that would work but it is worth considering.

cassimons (approved):

MattWellie:
@cassimons I've altered the logic a bit, so that there is an associated 'date of first discovery' with each category in the cumulative results. Currently this will just be stored in the historic results file, in future (i.e. when the report supports sorting/filtering within the results display) it will be easy enough to annotate each of the variants from _this_ AIP run with the date it was previously found instead of just doing a blanket removal of anything previously seen. - or doing both and writing two outputs. In either case, this PR will start writing date-tagged variant results into the results folder, and we can incorporate that however we like in future

cassimons:
Awesome thankyou!

PR #169, opened 2022-12-06
161 base jinja template for report

daniaki:
## Proposed Changes
  - Jinja2 templating with Bootstrap5 an Datatables

## Checklist
- [x] Related Issue created
- [x] Linting checks pass


MattWellie:
Couple of things to investigate -
Variant IDs aren't rendering as hyperlinks
CSQ field is first-character only

MattWellie:
contains changes to 30+ files because I'm a sucker for a messy rebase.
In addition to @daniaki 's templating changes, I've made a few additions - 

- Tooltips (thanks @danielreti); now the 'gold stars' for when a gene overlaps a panel relevant to a participant's phenotypes are now literal gold stars. Hover over for the panel name & version, and click on the star to be taken to the relevant page in PanelApp
- The panel table also contains a hover-over tooltip hyperlink, but displays as the panel name. Maybe this one should have been a simple hyperlink, it won't make much difference in practice though
- Some columns have been rearranged, to prioritise the MOI as one of the most important columns

This could be extended in a couple of ways, but I think it's a good stopping point for now. Next move is to get this whole application bundled into a docker image, so I'd like this change in prior to that

PR #173, opened 2022-12-12
Make site-only VCF before passing to VEP

vladsavelyev:
The problem: VEP jobs expect a site-only VCF, and would run out of disk when used with a full VCF with genotypes. 

Adding a job to strip off genotypes.

Depends on https://github.com/populationgenomics/production-pipelines/pull/227


MattWellie (commented):
damn you're fast

MattWellie (approved):

PR #180, opened 2022-12-23
Containerise (low priority)

### LINKED ISSUE #179: Make an AIP image, instead of running installs
So much time gets wasted running the same code over and over - copying and installing the whole codebase.

This was a stopgap measure that's gone on way too long.

Create a stable image, and use that as the job execution environment. 

Allow for development pushes to be built as images - not sure if this should be automatic, or using manual workflow invocations

MattWellie:
# Fixes

  - Closes #179

  - Creates a Dockerfile, a reasonably extensive `.dockerignore`, and a github yaml to run the container building process. 
  - Updates bumpversion conf to update the dockerfile contents
  - AFAIK this repo has secrets enabled ~but no runner to execute, so nothing will be built yet~ the runner(s) are obviously set up for CI, Matt is just an idiot

I'd like to discuss how `cpg_workflows` is built as a comparison - I'm happy either building only from main, or on deliberate invocations of the build action, rather than with every commit. 

I've added a secondary build file that will build an explicitly `test` version of the image, and will only run when manually invoked

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


vladsavelyev:
Looks good! Manual dispatch for test builds is a good alternative (automated builds on every commit like for cpg_workflows also works good).

There is a lot of documentation and formatting change in this PR besides dockerisation, is this intentional? Would be easier to split documentation and code changes into separate PRs.

MattWellie:
> There is a lot of documentation and formatting change in this PR besides dockerisation, is this intentional?

Nope, that wasn't intentional! I've changed the base branch now (I'd like the docker changes and corresponding documentation changes to be in main at the same time... but not important)

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,48 @@
...
+      uses: google-github-actions/setup-gcloud@v0
+      with:
+        project_id: cpg-common
+
+    - name: gcloud docker auth
+      run: |
+        gcloud auth configure-docker australia-southeast1-docker.pkg.dev
+    - name: build
+      run: |
+        docker build . -f Dockerfile --tag $BASE_IMAGE:${{ github.sha }}
```
illusional, line 40: You could build this as whatever locally, not that it super matters anyway - but you're not restricted by the github.sha

vladsavelyev (approved):

PR #185, opened 2023-01-08
Fix genotype bug

MattWellie:
# Fixes

  - N/A

## Proposed Changes

  - there was a genotype display bug where both Het and Hom would display as `Het` - a silly conditional copy/paste error
  - unrelated to, but found because of that error there is also an MOI issue:

1. PanelApp has a whole range of MOI options, including niche variations like `X-LINKED: hemizygous mutation in males, monoallelic mutations in females may cause disease (may be less severe, later onset than males)`. I'm parsing these options into small Monoallelic/Biallelic/mono_and_biallelic instead of implementing an MOI test 1:1. For this example I'm reducing this to `X-linked Monoallelic` based on a conditional check flow.
2. Initially the PanelApp query JSON contained the full strings, and they were reduced later on to the 'simple' representation
3. Recently I moved the simplification logic so that the JSON already contains the simplified string
4. I forgot to update the downstream logic, so it was trying to re-simplify the simplified MOI string, e.g. 

```
PanelApp : BOTH monoallelic and biallelic, autosomal or pseudoautosomal
Conditional: "Starts with 'both', not on X"
Simplified : Mono_and_Biallelic

Stored: 'Mono_and_Biallelic'
Conditional: "starts with 'mono', not on X"
Simplified: Monoallelic
```

Pretty stupid, but AFAIK that's the reason POLG isn't being evaluated as both recessive and dominant

The fix here is to remove the re-simplification, and just use the contents of the PanelApp JSON directly

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/interpretation_runner.py
+++ b/reanalysis/interpretation_runner.py
@@ -56,7 +56,9 @@ PANELAPP_JSON_OUT = output_path(
 )
 
 # output of labelling task in Hail
-HAIL_VCF_OUT = output_path('hail_categorised.vcf.bgz')
+HAIL_VCF_OUT = output_path(
+    'hail_categorised.vcf.bgz', get_config()['buckets'].get('analysis_suffix')
+)
 
 
 # local script references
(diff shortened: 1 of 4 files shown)


MattWellie:
Also I'm writing the tiny VCFs to `-analysis`, because getting the raw data to debug this was a pain

cassimons (approved):
Good find!

PR #189, opened 2023-01-10
tag using version, not SHA. Parametrise test tag

MattWellie:
# Fixes

  - parametrises the test tag (which should allow the arg to be overwritten)
  - removed the github SHA from the image name - @illusional mentioned this on a prior PR, but I hadn't realised how massive that name was until the images started being pushed

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -37,12 +37,11 @@ jobs:
         gcloud auth configure-docker australia-southeast1-docker.pkg.dev
     - name: build
       run: |
-        docker build . -f Dockerfile --tag $BASE_IMAGE:${{ github.sha }}
+        docker build . -f Dockerfile --tag $BASE_IMAGE:$VERSION
 
     - name: push latest
       if: ${{ github.event_name != 'pull_request' }}
       run: |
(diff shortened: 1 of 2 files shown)


vladsavelyev:
> removed the github SHA from the image name - @illusional mentioned this on a prior PR, but I hadn't realised how massive that name was until the images started being pushed

Can you give more context on that? What do you mean that the name is massive? Can you give a pointer to Michael's comment?

vladsavelyev (approved):
Looks good!

Personally I find it handy to build SHA-tagged images for cpg-workflows as a dev tool, as it allows me for a quicker iteration on testing fixes before they are merged; but probably not that needed for AIP :)

MattWellie:
@vladsavelyev The comment I was thinking of is [this one](https://github.com/populationgenomics/automated-interpretation-pipeline/pull/180#discussion_r1059985847). I don't think this would affect the tags being used, as I'm explicitly re-tagging in the workflow, so that change is not consequential.

The main change here is in the docker_test workflow, as the name param wasn't used previously. I'd like to change this up so that I can name the test image after the branch it relates to, for example, so I can have a few different test images in parallel for specific test runs.

PR #205, opened 2023-01-17
test tag fix

MattWellie:
# Fixes

  - completely biffed on how to get the input variable into the actual tag when designing the test workflow
  - I think it's still important to permit multiple separate in-development branch images to be built, so I'm avoiding hard-coding 'test' as the tag

## Proposed Changes

  - moved to the corrected (?) syntax, modelled on [this](https://github.com/populationgenomics/analysis-runner/blob/main/.github/workflows/deploy_server.yaml)
 
## Checklist

- [ ] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/.github/workflows/test_docker.yaml
+++ b/.github/workflows/test_docker.yaml
@@ -37,4 +37,4 @@ jobs:
         docker build . -f Dockerfile --tag $BASE_IMAGE:{{ github.event.inputs.tag }}
     - name: push test
       run: |
-        docker push $BASE_IMAGE:$tag
+        docker push $BASE_IMAGE:{{ github.event.inputs.tag }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test_docker.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
         gcloud auth configure-docker australia-southeast1-docker.pkg.dev
     - name: build
       run: |
-        docker build . -f Dockerfile --tag $BASE_IMAGE:$tag
+        docker build . -f Dockerfile --tag $BASE_IMAGE:{{ github.event.inputs.tag }}
     - name: push test
       run: |
         docker push $BASE_IMAGE:$tag
```
illusional, line 40:
```suggestion
        docker push $BASE_IMAGE:{{ github.event.inputs.tag }}
```

illusional (changes requested):
Need to change the push too

.github/workflows/test_docker.yaml
MattWellie: good catch 😬

illusional (approved):

PR #211, opened 2023-01-19
Fix panel stuff

### LINKED ISSUE #210: Bug (?) in MOI application
Paul drew attention to an issue with how the MOI for a particular gene is being found -

1. Query for the Mendeliome (where we know each gene has a curated MOI)
2. Query for any subsequent panels
  * for each gene, collect all MOIs on all panels
3. Once we've finished collecting data, evaluate the collected MOIs and find the most lenient applicable MOI

This made sense as a way of combining the MOI data from multiple panels, but causes an issue:

## Blanket MOI test

```
Each gene is assessed at the most lenient panel-matched MOI, even if a specific
participant would not have had a more lenient MOI phenotype-matched to them
```

e.g. for [SLC22A5](https://panelapp.agha.umccr.org/panels/entities/SLC22A5) - the MOI is Biallellic on every panel except one. If the Cardiomyopathy_Paediatric panel is included in the analysis, this gene will always be assessed as `Monoallelic or Biallelic` for all participants.

This is a difficult one to account for, as by including a more lenient MOI we are conceding that this is a mode of pathogenicity - except we want to apply different MOI leniency to different participants.

## Separate

Additionally, an issue exists where the MOI 'other' as a placeholder is not being accounted for. We are recognising 'Unknown' and processing that explicitly, but 'Other' is not recognised, so inclusion of any gene with an 'other' MOI on any panel will be accepted as Monoallelic or Biallelic. This also appears to be a way to evade the `X-Linked` MOI detection:

```
PDHA1, X-Chromosome https://panelapp.agha.umccr.org/panels/entities/PDHA1
All MOIs are X-linked or Unknown except for Paroxysmal Dyskinesia: https://panelapp.agha.umccr.org/panels/259/, where it is Other

Paroxysmal Dyskinesia is included in the Dystonia_Superpanel, where it is `Other` (THERE'S ANOTHER ISSUE HERE, STICK A PIN IN THIS ONE)

Because the `Other` MOI is not being processed appropriately, this X-Chrom gene being assigned the 'Mono_And_Biallelic' MOI and is being processed using autosomal MOI (i.e. variants will not be excluded based on hemi count in Gnomad
```

## And Finally

PanelApp panels can be nominated as 'SuperPanels', including multiple more granular panels. The lesser-included panels logic in PanelApp is causing a problem - an example is the Dystonia Superpanel 3123.

This panel https://panelapp.agha.umccr.org/panels/3123/ [API response](https://panelapp.agha.umccr.org/api/v1/panels/3123) contains PDHA1 twice, included from 2 different sub-panels. On each of those it has a different MOI. That wasn't something I expected in the script...

# Solutions?!

1. Ignore all `Unknown` & `Other` MOI - decide only on the basis of populated MOIs, defaulting to 'Mono_And_Biallelic' only if there were no populated MOIs at all. This should solve the 2nd point above.
2. Discuss the participant-specific MOI match with analysts. My gut feeling is that the logic gets a lot chunkier (see: buggy) if we move to a per-participant MOI test so I would like to avoid that if possible. For now we can have a 'disclaimer' saying that MOI will be processed at the most lenient level on any phenotype matched panel, and this can be revisited in future.
3. Identify superpanel discrepancies as targets for curation?

## Side note

Make the list of panels a set, there's tons of duplication there due to re-reading the 'historic' data

MattWellie:
# Fixes

  - Closes #210

## Proposed Changes

  - Many
  - The PanelApp MOI simplifier takes the Chromosome for the gene into account, correctly identifying the default MOI for sex-chromosome genes.
  - Alters the MOI parsing from PanelApp so that `unknown` and `other` are ignored. When generating the summary MOI for a gene, continue to take the most permissive of the 'real' MOIs found across all assessed panels. If there were no MOIs (i.e. every MOI was `unknown/other`), default to taking the most lenient possible for the chromosome. Previously `other` was not accounted for, so an MOI of `other` automatically tricked the logic into applying the most lenient MOI - `mono and biallelic, autosomal` - even when the gene was on chrX.
  - Swaps out the simple MOI parsing for a match/case statement (new in python 3.10 and really sexy).
  - Converts the list of panel IDs associated with a gene into a set. Previously panel IDs were appended to a list, and could incur duplicates. Due to the way the 'historic' data is cumulatively built, we could end up with absurdly large datasets if this isn't cut down to unique only.
  - Added various chromosomes to the panelapp test fixtures, and enforced all the panel IDs being integers.
  - Updated the canonical 'pre-panelapp' JSON to have all PanelIDs as integers (this housekeeping change alone is ~19000 lines).

## Discussion?

I don't have a simple fix for the `Blanket MOI test` issue on #210 (if one participant is phenotype-matched to a panel with a very lenient MOI, that will apply to the whole cohort).

```
... CLCNKB and SLC22A5 for example are matching to a dominant MoI in the AIP, 
where they are biallelic in the Mendeliome
```

I'm unsure if this is an issue worth solving:

- From [this thread](https://centrepopgen.slack.com/archives/C035QN8C0DN/p1674015797191739) I gathered (rightly or wrongly) that the Mendeliome panel should contain the most lenient MOI under which a gene can cause a disease, and the Mendeliome is still forming the spine of AIP analyses. By first querying for the Mendeliome contents, we're expecting to find the most lenient applicable MOI, making this discussion moot for most genes.
- Assuming cohorts are clustered by phenotype or disease category, the logic of applying the same MOI for a gene to all members of a cohort is pretty reasonable, even if a family wasn't personally matched to a lenient MOI.
- The only terrible scenario here was when real MOIs were overruled by the inclusion of a less well curated panel where the MOI is marked 'other' - this previously meant the gene was assessed under the most lenient possible MOI, but that buggy case has been treated by the changes above.

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/utils.py
```diff
@@ -607,51 +606,46 @@ def write_output_json(output_path: str, object_to_write: dict):
...
-        if 'biallelic' in panel_app_moi:
-            simple_moi = 'Hemi_Bi_In_Female'
-        else:
-            simple_moi = 'Hemi_Mono_In_Female'
-
-    return simple_moi
+    if input_moi in IRRELEVANT_MOI:
+        raise ValueError("unknown and other shouldn't reach this method")
+
+    default = 'Hemi_Mono_In_Female' if chrom in X_CHROMOSOME else 'Mono_And_Biallelic'
```
cassimons, line 628: Am I correct that this will be applied if there is no proper MOI in any of the panels used? If so, I am worried that this will end up being too verbose. It may be better to spit our a "no MOI" warning then default to:
```Hemi_Bi_In_Female' if chrom in X_CHROMOSOME else 'Biallelic'```

cassimons (approved):
I *think* the fall back default MOI should probably be `Hemi_Bi_In_Female' if chrom in X_CHROMOSOME else 'Biallelic'`  - other than that this looks great.

reanalysis/utils.py
MattWellie: Updated this PR to use biallelic/Hemi and Biallelic as the base MOI. The default could also be taken from the config, but I don't think we'll be coming across this situation that frequently.

PR #215, opened 2023-01-24
remove tons of unused MT data up front - shrink checkpoint times

### LINKED ISSUE #214: Clinvar Path missing from report
Context: "Any chance you could find out what stage [this variant](https://seqr.populationgenomics.org.au/project/R0021_mito_disease/saved_variants/variant/SV0000659_1545069018_f000316_a) is being kicked out of the [AIP report](https://main-web.populationgenomics.org.au/mito-disease/reanalysis/2023-01-16/summary_output.html)?"

Investigated, result: 

The cohort AC for this call is `29` from an AN of `2718`. This results in an internal frequency in this callset of just over 1%, so the variant is removed.

Proposal - alter the common-in-joint-call filter to allow any Clinvar Pathogenic variants through, mirroring the approach taken with the Gnomad AF filter.

Separate issue: Seqr is showing the joint-call AC as 27, instead of 29 from the Matrix Table. Q - is this an artefact of the variant annotations being saved when it was originally tagged, so the AC is not accurate?

MattWellie:
# Fixes

  - Closes #214
  - Clinvar known pathogenic variants are removed from the dataset if they are over 1% within the cohort

## Proposed Changes

  - Disables both joint-call and population AF filters where a variant is known pathogenic
  - We already defer to Clinvar ratings over population frequencies, this extends that to the within-cohort joint-call frequency
  - The specific variant that started this was [this one](https://seqr.populationgenomics.org.au/project/R0021_mito_disease/saved_variants/variant/SV0000659_1545069018_f000316_a), which is now found using an image built from this branch [here.html](https://test-web.populationgenomics.org.au/mito-disease/reanalysis/chr15_2023-01-24/summary_output.html)

## Bonus unrelated change

  - scraps all annotations not required to run AIP prior to the first checkpoint. This is an effort to reduce the time taken to run the workflow by discarding all data which is not referenced in any decisions, and will not make it into the VCF
  - The current implementation executes `checkpoint_and_repartition` 3 times, so reducing the volume of written data by... a lot? half? 70%? Should massively reduce runtimes. To be tested.

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/.github/workflows/test_docker.yaml
+++ b/.github/workflows/test_docker.yaml
@@ -34,7 +34,7 @@ jobs:
         gcloud auth configure-docker australia-southeast1-docker.pkg.dev
     - name: build
       run: |
-        docker build . -f Dockerfile --tag $BASE_IMAGE:{{ github.event.inputs.tag }}
+        docker build . -f Dockerfile --tag $BASE_IMAGE:${{ github.event.inputs.tag }}
     - name: push test
       run: |
-        docker push $BASE_IMAGE:{{ github.event.inputs.tag }}
+        docker push $BASE_IMAGE:${{ github.event.inputs.tag }}
--- a/reanalysis/hail_audit.py
+++ b/reanalysis/hail_audit.py
@@ -10,6 +10,103 @@ import logging
 import hail as hl
 
 
+BASE_FIELDS_REQUIRED = [
+    ('locus', hl.LocusExpression),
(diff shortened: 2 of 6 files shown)


MattWellie:
Also contains a lot of neutral docstring changes

cassimons (approved):

PR #217, opened 2023-01-27
new in prev. 12 months

### LINKED ISSUE #216: Remove strict reliance on optional files
# Issue

The per-participant panels and 'new' genes inputs are required in the current format, even though they should be optional

---

### Problem

A JSON describing the panels to use per participant cannot be None, else the final stage will fail. 

### Solution

Allow this file to be missing, and if so, don't do any gene-list filtering (with no personalised listings, all genes used are relevant to all families). 

Additionally, allow individual participants to be missing from that lookup, in which case only the default panel for the analysis will be used. 

---

### Problem

AIP can use a nominated folder to store data from previous runs. Panel details for each run are cached to this folder. With each run the current panel details are compared to the most recent previous run to determine which genes are 'new'. If this file is missing completely, all genes are treated as new.

### Solution

Permit this to be missing, in which case don't treat any genes as being 'new' - the alternate case is to treat everything as new, which is crap. 

### Discussion?

The 'pre-panelapp.json' file bundled with this codebase represents a VCGS-specific implementation of when genes should be 'new', that is not expected to apply to all other deployments. Instead of using this to determine when a gene is new, don't use the new mechanic at all. 

An alternative would be to always use this VCGS data, or to choose a specific version of the mendeliome as a default base.

### LINKED ISSUE #218: Removal-From-Panel Bug
Due to the way we are now aggregating data from previous runs, we can have unwanted spillage from prior runs. For Panel Data this 'prior data' will contain each gene previously seen for a given cohort and all panels it has been seen on.

```
{
    ENSG1243: {
        panels: [
            1, 
            12, 
            123
        ]
...
```

We use this list of prior panels to [check if the gene is `new`](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/reanalysis/query_panelapp.py#L110-L111) on a panel, and then there's a bug where the [full list of panels](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/reanalysis/query_panelapp.py#L137) is annotated into the current run's data, even if it is **no longer present** on a panel. 

```
if gene X was on panel A and B, the prior data will contain both A and B as prior panels.

If gene X is no longer associated with panel A, curators will remove it.

A new Query run will find gene X on Panel B, but the prior data will lead to both A and B being associated.

If a participant is phenotype-matched to Panel A, and they have variants in Gene X, those will be 
inappropriately shown in the report, even though curators have removed the association
``` 

This should be solve-able by better segregating the old & new data. The cause of this is lazily building all current and prior data into the current object, then saving that as the new historic data.

I feel like the PanelApp query script is sagging under too many ideas being glued on top of each other, and could do with a refactor

## Solution

* Separate out the Current run's data from the Historic data
* Historic should accumulate all panels for each gene
* Current run should take only data from the current run's panels

MattWellie:
# Fixes

  - Closes #216
  - Closes #218
  - Panels json non essential

## Proposed Changes

  - In the absence of a 'previous' gene list or historic data, Cat. 2 uses a diff between now and 12 months ago as a default behaviour (# months configurable)
  - simplifies the historic format down to `{GENE: [list, of panels], }`
  - simplifies the VCGS pre-panelapp data to reflect this
  - 

## Notes

Querying now for the Mendeliome now vs. 12 months ago only results in 7 new genes

Comparison with now vs. the pre-panelapp mendeliome - 681 new genes

Not a like-for-like comparison, just leaving that as a note

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


reanalysis/query_panelapp.py
```diff
@@ -189,6 +196,84 @@ def read_panels_from_participant_file(panel_json: str) -> set[int]:
...
+    iterate through its associated activities
+    return the panel version which was closest to 12 months ago
+
+    or return None, i.e. if the panel is not >= 12 months old
+
+    Returns:
+        a version string from 12 months prior
+    """
+
+    date_threshold = datetime.today() - relativedelta(years=1)
```
cassimons, line 211: Can we (now or in the future) make this configurable to be either X months old or from a specific date?  I am not worried if you add the logic now or merge as is then include this down the track.

cassimons (approved):
LGTM.  If we can make the time delta configurable at some point that would be grand - but that can wait.

reanalysis/query_panelapp.py
MattWellie: Doable! I'll stick something in the conf to have 'months' - `relativedelta` works fine with any number of months, not capped at 12

MattWellie:
Test run ongoing
https://batch.hail.populationgenomics.org.au/batches/420854

MattWellie:
Fixes a bug with the prior match-case statement where "monoallelic, autosomal..." wouldn't match to "monoallelic" because of the comma. Now includes a complete punctuation strip prior to the split

PR #222, opened 2023-02-02
Lenient de novo toggle

### LINKED ISSUE #171: Split male and female X-inheritance into two separate methods
Currently the X-Chrom inheritance models are functional, but not very modifiable. There's so many conditional clauses that the code is tough reading, and will be tough to maintain if required. 

Instead, fragment these into a Male and a Female version - e.g. the X-Recessive MOI model will be the result of running [x-recessive male, x-recessive female]

That would negate the need to do sequential checks for male and female participants, or for bunching together male het & hom with female hom, etc.

### LINKED ISSUE #194: Insufficient Flexibility RE: variant caller specific fields
Slack with Miah - e.g. the Phasing field [`PS` we're using](https://github.com/brentp/cyvcf2/issues/31) is alternatively present as `PGT` and `PID` in [GATK](https://gatk.broadinstitute.org/hc/en-us/community/posts/360057608491-What-is-the-physical-phasing-)

Also, try and find out why we're generating this as PS - is it to do with Picard relationship testing as part of the pipeline?

### LINKED ISSUE #221: Sneak in home-brewed Prob. Likelihoods where missing
The de novo functionality we are using requires the presence of PL scores with each variant call, especially the WT calls. For at least one collaborator this is missing, as the data is cleaned down to GQ only, to reduce the data footprint.

Discussion - for the limited purpose of the de novo finder calculation, where we detect that PL scores are missing, should we substitute our own based on the derived GQ?

Doesn't feel robustly scientific...

As a thought experiment, we could check in an arbitrary variant whether the WT calls have associated PL values - if they are _missing_, replace PL entries across the whole callset with a GQ-linked dummy value;

```
# take the first variant
mt1 = mt.head(1)

# filter to real homref calls (i.e. not ./. missing)
mt1 = mt1.filter_entries(mt1.GT.is_hom_ref())

# pull out all the PL values and check for missing
do_replacement = None in mt1.entries().PL.collect()

if do_replacement:
    blah blah blah
```

we can do a case-replacement, e.g. having stand-in replacement values for GQ 0-10, 11-20, 21-30..., so the PLs added in aren't a complete fabrication.

Just a thought on how this could work if we want to rescue de novos from missing PL scores

### LINKED ISSUE #223: Make Support Great Again
Variants which are only assigned 'CategorySupport' are sidelined in the report, only showing up as a hyperlink in an annotation column. 

Alter the MOI logic so that both sides of a comp het always show up on the report independently.

do this without undermining our aversion to giving too much weight to support variants (e.g. don't accept them under dominant MOIs, or under a recessive MOI as Homozygotes)

### LINKED ISSUE #224: Cat. 2 issue
MOI tests use delayed filtering to reduce Cat 2 (new gene-disease assc.) variants

When Cat. 2 flags are assigned to a variant, this is done cohort-wide. Once successful MOIs are found, these are back-filtered: `If this gene is not new for this participant, i.e. not new one one of the phenotype matched panels, remove the cat 2 flag`

This runs into an issue as the partner variant is not checked in comp-het scenarios. e.g. 

```
Var 1 and Var 2 are in the same gene, marked as Cat 2

Var 1 is also a Cat 1

These variants form a reciprocal comp-het, so they are added as 2 events, one in each orientation

When we reach the filtering stage, assuming that the gene is not new *for this participant*:

The comp-het with Var 1 as a primary will have the Cat 2 stripped off, and still has Cat1 assigned, so it is retained (even though Var 2 should not have any assigned category)

The comp-het with Var 2 as a primary has the Cat 2 stripped off, and no category remains on the primary variant, so it is dropped.
```

This could be remedied with a fairly complex re-casting of the Cat 2 annotation as a 'sample' category:

- During the hail stage, read the panelapp and phenotype-matched panel data
- Assemble a gene-indexed dictionary where the values are all samples for which this gene is new
- Annotated the sample IDs as the Cat 2 label instead of a boolean value that has to be double checked later.

For large cohorts that might not be a great solution... If a new gene is added in a core panel, that label could contain thousands of sample IDs compressed into a string. Bad. Also Hail dictionaries can take ages to build and query.

Alternate: When reading the variant from the VCF, translate the Cat 2 field into this structure instead of just keeping a boolean. For each variant:

- if the sample has a Cat 2 flag:
- find all samples IDs which have at least one alt allele and are affected
- find out which of these samples (if any) this gene will be 'new' for
- store the category 2 label as 'categorysample2', containing only the samples where they could be considered a participant and this is a new gene

To be completely honest I'm not sure this situation has arisen yet, but if it does it will be an issue

MattWellie:
# Fixes

  - Closes #221
  - Closes #194 
  - Closes #171
  - Closes #223 
  - Closes #224 - Instead of the previous approach of passing all variants, then filtering back to what is appropriate for each sample, I'm manipulating the Cat2 (new) flag. In Hail this remains a cohort-wide flag, binary. In the MOI testing phase we check which panels the relevant gene is new in. If the gene is new in the 'core' panel (Mendeliome), this remains a boolean, as it applies to all samples. If it is new in a phenotype-matched panel, the categoryboolean2 flag becomes a categorysample2 flag containing the samples for which that gene should be treated as new. That way, when we test 'is this variant classified for this sample', we accept either the boolean or the sample approach with no further downstream changes.

## Proposed Changes

  - Throws in a few ideas, not sure all of them need to make it in...
  - Changes the base build for the Dockerfile, this is required, but not linked to a specific PR
  - Adds a new config entry - `conf:filter:lenient_de_novo`. I tried subbing in the super-naïve de novo calculation, but it's just not very good... Instead, this will insert mocked PL values for any HomRef calls, which lets the Hail builtin de novo caller work  on the RGP dataset. That behaviour means that the 'lenient_de_novo' name tag is probably not appropriate... Before tweaking the config key name it's probably best to decide if we would ever want this, or if the mock PL inserted is actually responsive to the GQ value, e.g. having different mock PLs to insert which actually represent a realistic value at GQ < 10, 10-20, 20-30... which would still generate realistic results from the involved de novo calling process
  - Will we ever use this other than through the externally called RGP data? Is this going to be useful for MSFT/Broad to run on their own datasets?
  - Reverts the sites only VCF method to something that works 🙃 - **maybe revisit this**
  - Phase logic works with PID/PGT, instead of only working with PS; this solves another issue with the RGP data
  - Submitters to be filtered out of Clinvar entires now altered by config - e.g. if we want to remove Broad submissions from this dataset
  - Alters Comp-Het test to allow a supporting variant to pass through if it pairs with a 'full' category - this means both will be represented in full in the report
  - Splits out MOI tests to better compartmentalise the logic (i.e. XDominant, XRecessiveMale, XFemaleHom, XRecessiveFemale are all split out into separate methods, each concerning only the relevant model). This means that the filtering logic can be less absurd. Still absurd, just less so. Likely to have some segregation issues, e.g. if a variant is hemi in a male and Hom in an unaffected female, the segregation will no longer spot this.
  - Massive changes to the Cat2 logic (see `Fixes`)
 
## Checklist

- [x] Related Issue created
- [x] Linting checks pass


cassimons (commented):
Copy pasta from slack:
Looking at the PR it seems you are substituting in a fixed PL of `[0, 200, 1000]`. Reminding myself of the GQ formula, would it be feasible to substitute that for `[0, $GQ, 1000]` ?
I think this would then be very close to what we would have if we had the original PLs (with a slight penalty from the GQ capping at 99).

cassimons (approved):

PR #234, opened 2023-02-12
release command correction

MattWellie:
# Fixes

  - The last command was wrong

## Proposed Changes

  - The new command is right


lgruen (approved):

release_scripts/broad_rgp_release.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-gcloud storage --project=broad-rgp cp --recursive \
+gcloud storage --billing-project broad-rgp cp --recursive \
```
lgruen, line 3: In case this doesn't work, you might have to move the `--billing-project` before `storage`.

PR #242, opened 2023-02-21
first draft - really ugly

MattWellie:
# Fixes

  - There's no centralised store for links to latest AIP reports
  - Currently we're getting by using a pinned Slack message, but that's not ideal

## Proposed Changes

  - A script to crawl relevant dataset buckets, finding the latest HTML files in the 4 possible categories - Exome/Genome, Singleton/Regular (some cohorts may be singletons anyway, this relates only to the test cohorts where we've manually removed family structure for testing purposes)
  - This can be set to run on a regular basis with Cloud Scheduler, overwriting a single file that we can link to a `popgen.rocks` URL, or similar static address.
  - Now includes a bit of light Jinja templating

## Issues

  - During the recent migration to AutoClass bucket management, all the timestamps seem to be as of time of copy, rather than containing their true creation/update timestamps. This will normalise over time.
  - Cohorts to include in the checks are hard coded, will require updates as new cohorts are added. Due to the templating, if a cohort has no found reports it will have no presence in the index.

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


helpers/report_hunter.py
```diff
@@ -0,0 +1,109 @@
...
+
+from cpg_utils import to_path
+
+
+JINJA_TEMPLATE_DIR = Path(__file__).absolute().parent / 'templates'
+TEMPLATE = 'gs://cpg-{dataset}-main-web/reanalysis'
+WEB_TEMPLATE = 'https://main-web.populationgenomics.org.au'
+
+
+cohorts = [
```
cassimons, line 22: If we can pull this list from the master config when you have it that would be grand.

cassimons (approved):

PR #244, opened 2023-02-22
always do replacement for missing PLs

MattWellie:
# Fixes

  - The initial de novo identification approach doesn't work well with re-blocked VCFs due to the loss of some individual WT call qualities
  - After discussion with Kaitlin S, it sounds like the 'hacky' implementation here will actually do exactly the right thing, so long as it is based on the quality scores stored with the Genotype

## Proposed Changes

  - Prior to running de novo checks, identify all genotypes which are called WT (i.e. not missing), do have GQ scores, and lack PL scores
  - Replace the missing PL scores with an array generated based on the PL score - `[0, GQ, 1000]`
  - Remove this as an optional/switched behaviour

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


cassimons (approved):
Good call in this.

PR #246, opened 2023-02-23
Forced panels into report

### LINKED ISSUE #243: Cat 2 broken in report
The latest change which rearranges Cat 2 from a boolean to a sample list breaks the reporting printer

When we make a minified variant, transform all sample fields to boolean fields (because it's done WRT a specific sample)

MattWellie:
# Fixes

  - there is currently no way to force a panel onto a whole cohort (as opposed to phenotype-matching or altering the core panel)
  - Closes #243 

## Proposed Changes

  - adds an optional section to the cohort portion of the config file for a list of panel IDs to apply to the whole cohort
  - during panelapp parsing, first get panels based on phenotype matching, then add on any extra panels as appropriate
  - during gene-panel filtering and determination of when 'new' applies to a specific participant, use this panel list; i.e. if a gene is not new in a panel phenotype-matched a participant, but is new in a panel applied to the cohort, treat as new for all participants
  - in the final report recognise these cohort-forced panels, and identify with a symbol similar to the phenotype-matching
  - streamlined a couple of helper methods in the AbstractVariant model
  - updated some test fixtures (VCFs with unused annotations)

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


cassimons (approved):
🔥

MattWellie:
There's a lingering issue here with forced-match panels always showing up as phenotype matched. Gonna try to squash that

PR #249, opened 2023-02-28
Run report hunter

MattWellie:
## Proposed Changes

  - adds a workflow trigger with Cron tab to set off the report crawler/index builder
  - pull cohorts to check from metamist

### DIFF
--- a/.github/workflows/report_agg.yaml
+++ b/.github/workflows/report_agg.yaml
@@ -1,0 +1,35 @@
+name: 'report index generator'
+
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: '42 13 * * 1-5'
+
+jobs:
+  build_report_index:
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - name: "Submit report crawler"
+
+      - id: "google-cloud-auth"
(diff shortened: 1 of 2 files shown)


.github/workflows/report_agg.yaml
```diff
@@ -0,0 +1,35 @@
+name: 'report index generator'
+
+on:
+  workflow_dispatch:
```
illusional, line 4:
```suggestion
  workflow_dispatch: {}
```

illusional (approved):

helpers/report_hunter.py
```diff
@@ -38,16 +39,19 @@ def main():
...
-        bucket_name, prefix = root.removeprefix('gs://').split('/', maxsplit=1)
-        client = storage.Client()
-        blobs = sorted(
-            client.list_blobs(bucket_name, prefix=prefix), key=lambda x: x.updated
-        )
+    for cohort in ProjectApi().get_my_projects():
+
+        try:
+            root = TEMPLATE.format(dataset=cohort)
+            bucket_name, prefix = root.removeprefix('gs://').split('/', maxsplit=1)
```
illusional, line 48: You should use the template provided by the storage config for portability

PR #255, opened 2023-03-01
pass date to clinvar summariser

MattWellie:
# Fixes

  - date option was not passed in the clinvar wrapper
  - previous change somehow turned a TV on, this will rectify that mistake

## Proposed Changes

  - ... passes a date argument, if supplied
  - turns off the TV


vivbak (approved):
HAHAHHAHAHA 👀

PR #263, opened 2023-03-09
Annotate common tx csq

### LINKED ISSUE #140: Use same-codon clinvar annotations
- As an analyst 
- I would like variants to be annotated with clinvar consequences at the same codon
- So that I can see relevant consequences at each clinvar locus even if a specific change wasn't submitted before

When labelling variants with clinvar consequences, we would like to take into account pathogenic clinvar entries at the same codon. This can provide insight about possible consequences where a specific change doesn't appear in clinvar, assisting the curation process with additional context.

AIP categories 1, 2, and 3 are assigned partially on clinvar annotation. This will mean that these categories are assigned a little more generously on the basis of "clinvar is pathogenic". For this limited purpose we are not including benign annotation in the same codon, though all clinvar links to the same codon should be provided in the report.

1. Annotate each protein altering variant with other clinvar consequences in this codon
2. When selecting for 'clinvar pathogenic', allow for other pathogenic changes at this codon to count equally
3. In the exported annotations, include **all** clinvar alleles at this codon
    - i.e. if a variant has a clinvar benign and clinvar pathogenic change in the same codon, provide links to both in the report

MattWellie:
# Fixes

  - Implements a new category based on the PM5 ACMG criteria
  - Closes #140

## Proposed Changes

  - Includes a helper script with code used to take a clinvar VCF annotated by VEP, reshuffle the content into a Hail Table linking all Protein_ID and residue number with all relevant clinvar alleles
  - This helper isn't being built up into a proper wrapper script because once VEP is released on Hail Query that will be much easier to develop. Pretty manual until then
  - Includes a `categorydetailspm5` which will collect all clinvar missense submissions at the same residue as a current missense
  - any matched submissions that are an exact match to this current variant are removed
  - This data is processed to create a new column in the report HTML with hyperlinks to the other entries (if appropriate)
  - needs further testing, but happy to force this through for now to freeze functional code - happy to increase test coverage

Waiting for tweaks to downgrade the pm5 from being a full category to being a type of support

## Checklist

- [x] Related Issue created
- [] Tests covering new change
- [x] Linting checks pass

### DIFF
--- a/reanalysis/clinvar_by_codon.py
+++ b/reanalysis/clinvar_by_codon.py
@@ -1,12 +1,26 @@
 """
 method file for re-sorting clinvar annotations by codon
 utilises the latest in-house clinvar summary data
+this should probably be transformed into a broader wrapper;
+
+1. download ClinVar VCF
+2. annotate that VCF with VEP consequences
+3. substitute out the clinvar decisions with CPG versions
+4. munge the data to index all CPG decisions on protein change
+5. write resulting table out
+
+The main reason not to do this just yet is that we won't need
+to re-run it frequently, and hopefully it won't be long before
+VEP-on-Hail-Query works, in which case that whole workflow
+would be a trivial extension from this
 """
+
+
 import logging
 
 import hail as hl
 
-from reanalysis.hail_filter_and_label import PATHOGENIC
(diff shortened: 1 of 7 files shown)


cassimons (approved):
🔥

Issue #272, opened 2023-03-21
Autoclass is changing all the file modified dates


MattWellie:
Current Report Index page generation is utilising the Last Modified date on GCP files to find which files to present as 'latest'. The AutoClass migration seems to be ruining this concept... This Issue lines out a working hypothesis...

Every time an old file is downgraded to a different storage tier it is copied to a new location (??)
This creates a new edit date (??)
This results in older files being flagged as latest on the report

Still need to nail down the cause here.

Might transition to using the dates in file paths instead of using created/modified dates.

cassimons:
Ugg what a pain, and what bad semantics from gcp.

lgruen:
The original migration incurred a change of the "modified time", as we had to made a full copy of all bucket contents to change the bucket's class to Autoclass. However, as far as I understand this should have been a one-time thing -- i.e. that should _not_ happen for class migrations now that Autoclass is active...

MattWellie:
Dropping in an example here - I've tried to tiptoe around this as I believe loading the HTML page may re-locate the file into standard storage and reset all the dates...

https://console.cloud.google.com/storage/browser/cpg-ravenscroft-arch-main-web/reanalysis/2023-02-20/summary_output.html

This file has a created date of `17 Feb 2023, 10:04` matching the time of the AutoClass migration, and was since superseded by a new run on March 13th, which was picked up in the report index at the time. The Modified date of the file is `19 Mar 2023, 10:54`, and the storage class is Nearline so it has not been accessed in that time. That change came ~30 days after the migration took place, and the modified time was around Saturday at midnight, so must have been an automated change. 

If this is right and the modified date is being tweaked by AutoClass that's a bit irritating, but a clean solution is for AIP to be logging analysis outputs into metamist, and reading the latest reports the same way - I probably should have been doing that from the start so it's not a huge problem.

MattWellie:
I'm closing this as mitigated by the move to Metamist result logging

lgruen:
I'll ask Google tech support about this anyway.

PR #274, opened 2023-03-21
register html outputs in metamist

MattWellie:
# Fixes

  - AutoClass migration is playing havoc with 'find latest files by edit date'
  - The more robust process would be to log files in Metamist, retaining only one active file
  - Contributes to solving #272

## Proposed Changes

After completing the HTML generation process:
  - query for all 'web' files in the same dataset
  - find any which fit the same subtypes as the current analysis (genome/exome, familial/singleton)
  - If any active files are found, mark them as inactive (@illusional  I remember an issue where either querying for `active=True` didn't really limit the search, or updating active to inactive wasn't persisted. Is this still an issue?)
  - Register the current output file

This change would be a predicate for shifting the report index generator to using metamist.

The downstream plan would be to find the most recently created metamist entry, not necessarily the only active one - the 'deactivating' behaviour would be a nice to have for tidiness but not essential.

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/html_builder.py
+++ b/reanalysis/html_builder.py
@@ -17,6 +17,14 @@ from peddy.peddy import Ped
 
 from cpg_utils import to_path
 from cpg_utils.config import get_config
+
+from sample_metadata.apis import AnalysisApi
+from sample_metadata.model.analysis_type import AnalysisType
+from sample_metadata.model.analysis_model import AnalysisModel
+from sample_metadata.model.analysis_status import AnalysisStatus
+from sample_metadata.model.analysis_query_model import AnalysisQueryModel
+from sample_metadata.model.analysis_update_model import AnalysisUpdateModel
+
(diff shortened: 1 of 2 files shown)


reanalysis/html_builder.py
```diff
@@ -53,6 +61,70 @@ def variant_in_forbidden_gene(variant_dict, forbidden_genes):
...
+    Takes the output HTML from this analysis and registers it in
+    Metamist. Deprecates any existing HTML results
+
+    Args:
+        pedigree (str): path to the Pedigree file
+        html_path (str): path we wrote an HTMl file to
+    """
+
+    if get_config()['workflow']['access_level'] == 'test':
+        # never update metamist in test mode - no permission
```
illusional, line 75: You could access the test project, but there's probably not a lot of point

illusional (approved):
Looks good to me :)

PR #278, opened 2023-03-30
move metadata registration into a separate process

### LINKED ISSUE #45: Group Outputs by Families
Currently AIP generates per-sample results, where each affected individual across the whole cohort can be assessed independently (MOI tests are conducted per-family, but the operation is individual-centric). The output format is a dictionary, keyed by the sample IDs at the top level.

```python
{
    'sample_1': {
        'variant_1_key': _DETAILS_,
        ...
    },
    ...
}
```

This has not been a problem so far, as each family in the development dataset is a trio with a single affected participant. If we have multiple affected persons within a single family, we would expect repetition of results for each participant (under a complete penetrant model if MOI fits for one affected participant, it must fit for all affected participants).

In the JSON and HTML results each sample has a separate variant table, and the cohort-level stats count the number of variants per individual. This is potentially misleading, as we would be double-counting for every family with multiple affected persons.

Seqr links generated are per-family, so the same variant appearing against multiple family members is duplication of the exact same link.

```
If we move to analysing partial penetrance, is it possible that variants will pass MOI tests for some
affected participants and not others? I don't think so... it's the opposite scenario (presence in 
non-affected persons). So long as we are always treating affection status within a single family as 
relating to the same disease, we should no have any issues with this.
```

# Proposal:

Using the Pedigree, aggregate the results for all members of a family when 'simplifying' the results (removing redundant entries). Instead of presenting per-sample, we should present results per-family. 

- which family ID should we use? Seqr/External/Internal?
- should the JSON remain as it currently is, and only the HTML be simplified?
  - actually quite important... the comparison results are on a per-sample basis, so we would need that structure?

This requires a bit more thought so as not to break the interface with the comparison process

### LINKED ISSUE #270: Implement a CHANGELOG for alterations after 1.0.0
Find a changelog structure that we can use for documenting changes following the current feature-freeze. Items to identify:

- Interface changes (arguments, config, file formats, outputs)
- Behaviour changes (category addition/modification, MOI implementation)
- Appearance (new/changed features in the report)
- Documentation
- Bugs fixed

### LINKED ISSUE #271: Steal Miah's variant tagging functionality
The MSFT fork of this project includes some spicy tagging behaviour - a file containing samples, variants, and relevant tags can be supplied when generating the report. This will create tags alongside the relevant variants if published in the report, or it will collect any un-matched variants and tags in the run metadata tab. It's a good feature. Steal it. 

https://github.com/jeremiahwander/automated-interpretation-pipeline

### LINKED ISSUE #277: Restructure the results/presentation job
Problem: 

- Before a Docker image was established, containing all libs to run the results and report HTML, I was doing a filthy pip install of the local code and dependencies in the runtime container. To make this less horrible I bundled the results and HTML generation into the same container runtime to cut down on the number of separate filthy installs that were required.
- This means that although we treat the HTML generation script as a CPG-specific version, it's coupled tightly to getting any results at all
- The recent addition of a `sample-metadata` dependency means that anyone without a metamist connection can't generate any results (or... they'll get results, then the job will fail)

Solution..?:

1. Separate the result generation and the HTML generation into different container runtimes
2. Allow for some flexibility in how the HTML script is selected (config)
3. Allow some flexibility in how CLI arguments are selected for the HTML generation script
4. Make the Metamist registration optional (and possibly an entirely separate container instead of tacked onto the HTML generation??)

MattWellie:
# Motivation

Addition of the mandatory metamist file registration was dumb, and will break the MSFT implementation.

Given that I'm planning to ditch work on this codebase for a while I'm rolling a few currently open issues into this single change.

## Fixes

  - Closes #277 (?)
  - Closes #271
  - Closes #270 
  - Closes #45

## Proposed Changes

  - Don't make metadata registration a mandatory part of the HTML generation
  - Break Result JSON, HTML, and Metadata registration into 3 separate processes
  - Allow for the possibility of alternative scripts for HTML/other output generation - allow each alternative script to have different inputs
    - send all the keyword arguments you want to the output generation job, the chosen script will choose the arguments relevant to it based on the mini config
  - Allow for the potential of multiple metadata registration tools, registering most/all of the new files created with this run
    - similar process, though all that's being provided is the list of files and the relevant pedigree, the service-relevant content will be defined in the script

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


MattWellie:
https://test-web.populationgenomics.org.au/ohmr3-mendelian/reanalysis/chr2_denovo_test/summary_output.html was generated using this form (limited to just chr2 variants in the latest `ohmr3-mendelian` joint call)

cassimons (approved):

jeremiahwander:
Thanks for the refactor, this is going to be really helpful on our end!

PR #285, opened 2023-04-20
Update data_model.py

### LINKED ISSUE #283: Make data_model.py compatible with hail batch
On line 333 in automated-interpretation-pipeline/reanalysis/data_model.py there is a local write to a json file that causes an error when running hail batch. Works fine when running hail locally. I think this would be solved by adding code to support cloud paths.

MattWellie:
# Fixes

  - Closes #283  (hopefully)

## Proposed Changes

  - The path a JSON file is written to is derived from the ultimate output path
  - To support use in cloud and local versions, this open/write needs to be able to adapt to both situations
  - This JSON file doesn't need to be persisted at all, but removing the possibility of write clashes is required to parallelise this behaviour in unit test fixtures

### DIFF
--- a/reanalysis/data_model.py
+++ b/reanalysis/data_model.py
@@ -31,6 +31,7 @@ from os.path import join
 from dataclasses import dataclass, field
 
 import hail as hl
+from cloudpathlib import AnyPath
 
 from reanalysis.utils import CustomEncoder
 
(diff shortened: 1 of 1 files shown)


MattWellie:
Might be worth having a broader conversation if this gains traction, about whether the data needs to be written prior to being returned, whether the JSON data is full disposable (i.e. literally write to `./tmp.json` within the container which will be discarded), or any other I/O considerations

cassimons (approved):

Issue #287, opened 2023-04-28
CPG annotation process isn't safe to use


MattWellie:
See #269

Firstly, it _is_ safe for the main pipeline to use:

1. Generate a joint-called dataset
2. Run VEP on multiallelic data
3. Apply VEP data to variant data indexed only on `Chr:Pos` (partial key)
4. Split multiallelic data out into separate variant rows (Hail's splitting function knows to correctly partition vep content)
5. ...
6. PROFIT!

The AIP re-annotation cycle starts from the post-split data generated by this pipeline; if two different alt alleles are present at the same locus, these are two different rows. Because multiple rows can have the same locus, the partial-key join between VEP and variant data is broken - this will lead to indexing clashes, and the annotation from one row will be wrongly applied to any other variants at the same site. 

This has been seen in re-annotated AIP data (though this feature is rarely used internally), and in the re-annotation of the ClinVar VCF.

The only solution I can think of to solve this is to copy the chunks of cpg_workflows code relating to annotation, and alter how the data is pieced back together. That shouldn't be as bad as it sounds - the current code includes a ton of conditionals and switches for applying VQSR, or running in Dataproc. This version should be relatively minimal.

Issue #281, opened 2023-04-13
Clinvar Missense Process


MattWellie:
Extend the current ClinVar re-summary to also include generation of the PM5 attributes

1. Download ClinVar submissions
2. Process them into a Hail table of fresh decisions
3. Dump that table out to a given path
4. Filter the table down to Pathogenic SNVs
5. Write that as a sites-only VCF
6. Annotate that VCF using the cpg_workflows methods
7. Process the annotated MT to get the lookup of per-codon pathogenic changes

Reason - The ClinVar VCF contains 1.9 million entries. Processing this down to Pathogenic only leaves tens of thousands, probably less. 

The annotation takes ages

The process currently includes annotating the whole of ClinVar, joining with the privately summarised annotations, filtering for pathogenics, filtering down to missense SNVs, then reassembling to match on codon. 

Starting with Pathogenic SNV only -> shorter annotation run time -> shorter data reassembly time -> much less wasteful

MattWellie:
MAKE IT SIMPLE YA DONKEY

Issue #269, opened 2023-03-16
Investigate dodgy ClinVar annotation


MattWellie:
See #268 for the process to create new ClinVar annotations

See Slack https://centrepopgen.slack.com/archives/C035QN8C0DN/p1678918786123119 for the issue

The annotation of the ClinVar alleles resulted in two different base substitutions receiving the same VEP annotation - this was only found out downstream when a Nonsense variant in ClinVar was annotated as PM5 on the basis that it was a missense change. 

Poking at the annotated ClinVar Hail Table, chr5:132588728 C>G and C>T were annotated with VEP, both got assigned the annotations from C>G. 

Obviously this would be bad, but in trying to recreate the issue and check what's going on I ran into the QC issues as stated in #268

Relevant for #287

Issue #268, opened 2023-03-16
Create a ClinVar annotation wrapper


MattWellie:
NB I've avoided doing this so far on the basis that the implementation of VEP in Hail will make this a pretty trivial process. 

Process:

- Check if this needs to be done (i.e., check whether there is already an annotated ClinVar table present)
- Download the latest VCF of ClinVar variants (https://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/clinvar.vcf.gz)
- QC! Or complain to ClinVar until this is fixed, IDK
  - At time of writing the latest ClinVar's GRCh38 VCF uses GRCh37 contigs (1, 2 instead of chr1, chr2), so GATK and Hail both reject this until it is reheadered.
  - At time of writing the ClinVar VCF _SHOULD NOT_ contain any bases other than ACGTN, but the current VCF includes a range of IUPAC ambiguity codes, e.g. `(chr)11	5226739	15163	AG	YT`.
- Annotate with VCF using the standard pipeline method(s)
- Gather annotated fragments into a Hail Table
- Process the resulting Hail Table to get the Codon indexed ClinVar missense table

PR #288, opened 2023-05-02
Re-work VEP content

MattWellie:
# Fixes

  - Relevant for #287 #269 #281 #268 

## Issue

  - The main pipeline's annotation code doesn't work for the AIP use-case
  - The annotation code in `cpg_workflows` expects multiallelic data, such that each locus only appears once per dataset.
  - VEP is executed on multiallelic data, joined with Locus as the only common key, **_then_** the variants are split to have a different alt allele on each 'row'. Hail accurately partitions the VEP annotations corresponding to each alt allele.
  - AIP imports from this annotation library and allows for variant data to be annotated/re-annotated. As this data is (in our use-case) already split into individual alt alleles, the Locus-only join can cause problems where all alt alleles at the same locus will have the same annotation applied to each, e.g. in this situation, annotation from one of these variants would be wrongly applied to all 3:

```
chr1:12345 C>T
chr1:12345 C>TGAGTAGATC
chr1:12345 C>A
```

## Proposed Changes

Takes all of the methods we use from the cpg_workflows library and duplicates into AIP with a few changes:

- No VQSR data applied
- VEP data is indexed and joined on both locus and alleles, meaning that it can safely annotate a split VCF/MT
- None of the computed Hail scripts fields are handled (e.g. sortedTranscriptConsequences) because AIP doesn't use them. This should make the runtime and data footprint much smaller
- No switches to enable running in dataproc, or sorting/tabix'ing an intermediate VCF

Being able to correctly annotate variants split out into individual alleles is a prerequisite for using the re-annotation feature, and for being able to generate new PM5/Annotated ClinVar datasets

## Checklist

- [x] Related Issue created
- [ ] Linting checks pass


cassimons:
Does this mean we will still be using the VEP data from the prod pipeline if it is available or well we need to re-run VEP for each dataset?

MattWellie:
> Does this mean we will still be using the VEP data from the prod pipeline if it is available

Yup! The first pass through is fine because GATK calls in a multi-allelic way and the process is designed to complement that. The issue only crops up because the [splitting](https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/query_modules/seqr_loader.py#L68-L74) is non-reversible, so any subsequent re-annotations are on differently formatted starting data.

MattWellie:
To check that this annotation implementation works, I've taken a subset of the `schr-neuro` dataset (chr17 only)

The original whole genome results are here:
https://main-web.populationgenomics.org.au/schr-neuro/reanalysis/2023-03-13/summary_output.html

The subset results are here:
https://test-web.populationgenomics.org.au/schr-neuro/reanalysis/2023-05-03/summary_output.html

all chr17 variants are presented the same, with identical annotation. I'd also like to re-run the Clinvar annotation to check that the original issue is solved. There's tons of variants which share alleles in that dataset, so it should be a good test case. The only issue is how dodgy the ClinVar VCF formatting is...

cassimons (approved):

PR #290, opened 2023-05-02
Label lenient cat 1

### LINKED ISSUE #289: Flag possibly too-permissive MOIs
We've heard some concerns regarding overly-lenient application of MOI tests, specifically around genes with 'both monoallelic and biallelic' MOI in PanelApp.

Another contributing factor is how variants submitted to ClinVar do not have the MOI under which they were categorised as Pathogenic.

We would like to counterbalance this by tagging variants in the report with a warning when:

1. the MOI obtained from PanelApp is 'both monoallelic and biallelic'
2. the variant is flagged as Category 1 (ClinVar Pathogenic)
  - this applies when the variant has multiple flags, not limited to exclusively Cat. 1. As long as Cat.1 is amongst the applied categories the MOI testing is a little more permissive, and that's part of the problem

Not sure about the text to include in this label yet...

MattWellie:
# Fixes

  - Closes #289 
  
## Proposed Changes

  - If we are assessing a Cat. 1 variant with any of the most permissive MOIs (Mono_And_Biallelic or Unknown, and the Allosomal equivalent Hemi_Mono_in_Female) we always attach a default flag
  - Currently this is `Lenient MOI Used`, but happy to hear suggestions
  - The super lenient MOI(s) can be attached either because the Mendeliome/Base panel has that MOI, or as a result of multiple panels being used (Mono on one, Biallelic on another = Mono_And_Biallelic). This change will not differentiate between those two different situations. AFAIK the issue is the relationship between the ClinVar relaxed MOI testing and the lenient MOIs, so it doesn't really matter how we obtained that MOI.

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/moi_tests.py
+++ b/reanalysis/moi_tests.py
@@ -6,7 +6,6 @@ Reduce the PanelApp plain text MOI description into a few categories
 We then run a permissive MOI match for the variant
 """
 
-
 from abc import abstractmethod
 from copy import deepcopy
 
@@ -15,7 +14,6 @@ from peddy.peddy import Ped, PHENOTYPE
 from cpg_utils.config import get_config
 
 from reanalysis.utils import AbstractVariant, CompHetDict, ReportedVariant, X_CHROMOSOME
-
 
 # pylint: disable=too-many-lines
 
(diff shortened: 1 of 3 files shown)


MattWellie:
This isn't currently doing the right thing - we only want to apply this flag to variants which pass a **_dominant_** MOI check, in genes which permit both Dominant and Recessive MOIs

PR #293, opened 2023-05-08
Vep clinvar common

### LINKED ISSUE #268: Create a ClinVar annotation wrapper

### LINKED ISSUE #269: Investigate dodgy ClinVar annotation

### LINKED ISSUE #281: Clinvar Missense Process

### LINKED ISSUE #287: CPG annotation process isn't safe to use

MattWellie:
# Fixes

  - Closes #287
  - Closes #281 
  - Closes #269 
  - Closes #268 
  - Currently the ClinVar summary process runs per-month per-project, which is wasteful
  - Switches this out for a single cohort-agnostic CRON-triggered workflow which generates all ClinVar resources once per month

## Proposed Changes

  - Takes the annotation code from #288 and applies this to the task of generating an annotation table for the PM5 evidence category. This adds an additional option to the `annotate_cohort` method which is `vep_only` - with this boolean argument set to True the annotation process concludes after combining variant data with VEP results, as there is no need in this context to annotate with ClinVar or gnomAD etc..
  - This avoids an unresolved issue as stated in [ZULIP THREAD](https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/SocketException.20when.20writing.20Table) where the ClinVar data failed to complete through the annotation process.
  - The `clinvar_runner` script has been improved and altered, serving as an end to end workflow:

1. Localise the submission and variant files from NCBI's ClinVar FTP server
2. Obtain a Hail Table representing the internally re-summarised ClinVar alleles, save to a cross-project bucket ([see this](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/design_docs/Clinvar.md))
3. Filter the Hail table to SNVs only, and export the Hail Table as a sites-only VCF
4. Annotate this VCF using VEP, and combine those annotations back into a MatrixTable format
5. Spice up this MT by rearranging into a PM5 table (to be documented), suitable for use in AIP's Hail runtime

  - The end product of this run is a cross-project folder (`cpg-common-<main/test>-analysis/clinvar_aip/<date>/...`) containing the input files used, the decisions table and SNV VCF generated, the annotated form of that VCF as a MatrixTable, and a PM5 lookup table for use in an AIP run.
  - When AIP runs the two Hail tables are used as an annotation source

## Note 

I've re-drafted this so that the code will defer to the config values for `clinvar_decisions` and `clinvar_pm5`, so that it supports excluding submitters specific to a cohort.

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


MattWellie:
There are definitely more pressing priorities here, but this has been hanging around as a manually run task for months

cassimons (approved):

PR #295, opened 2023-05-11
updated clinvar workflow

MattWellie:
Can you pls check if this would be the right way to send extra config params via a curl?

failing batch as reference
https://batch.hail.populationgenomics.org.au/batches/424192/jobs/1


illusional (approved):
Yep, looks like it: 
https://github.com/populationgenomics/analysis-runner/blob/f83449b217bd09e0ad7dfe974abf964baa2b35f4/server/main.py#L124-L125

PR #314, opened 2023-08-23
Adds type and trio True/False to family metadata

### LINKED ISSUE #310: Add family size to family metadata?
Family-level metadata currently includes "ext_id", "family_id", "members", "phenotypes", "panel_ids" and "panel_names".  Could we add family_type and/or family_size info (perhaps aligned with categories in "family_breakdown" in the run metadata) to each family? This would make it easier to sort outputs based on family characteristics.

MattWellie:
# Fixes

  - Family metadata doesn't currently include an explicit count of family size, or a categorisation of the family type
  - Closes #310 

## Proposed Changes

  - Includes two attributes in the metadata for each participant:
    - family_size: count of number of participants in the Pedigree for this family
    - trio_present: True/False, True if the family contains a participant/mother/father trio
  - Thinking here is that a Trio being present/absent is a more useful separation than duo/trio/quad/etc.


MattWellie:
Closing this, needs to be reimplemented

PR #317, opened 2023-10-09
Remove solved cases

### LINKED ISSUE #316: Whittle out solved families
During run setup we need a way to remove families/probands which have been solved, so as not to clutter future reports. Undecided on how this should be provided:

- Via Config? - this is probably the quickest and easiest, but requires the same config file being kept between runs
- Via tag against a participant in metamist?
- Via something else? (Exported Seqr tag?)

And how should this manifest?
- Complete removal from the analysis (i.e. not in pedigree at all)
  - remove all solved participants
  - remove all participants with no remaining affected relatives
- Recorded removal (i.e. the report should identify participants who were not considered on the basis of a prior solve)

MattWellie:
# Fixes

  - Closes #316 

## Proposed Changes

  - Adds an optional argument `--solved` to pass through a list of participant IDs
  - These IDs are passed through to the MOI testing process, where they are marked `solved: DATE` in the results JSON (results are processed and retained throughout)
  - In the HTML generation step, these solved cases are removed from the presentation table, and added to the Statistics tab with `ID: DATE`
  - The solved cases are recorded in the history/state file for the project, so that adding a participant as `solved` will result in them being hidden in all future analyses

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/html_builder.py
+++ b/reanalysis/html_builder.py
@@ -125,7 +125,13 @@ class HTMLBuilder:
 
         # Process samples and variants
         self.samples = []
+        self.solves = []
         for sample, content in results_dict['results'].items():
+            # remove solved cases
+            if solve_date := content['metadata'].get('solved'):
+                self.solves.append({'sample': sample, 'solve_date': solve_date})
+                continue
+
             self.samples.append(
                 Sample(
                     name=sample,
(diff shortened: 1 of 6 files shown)


MattWellie:
Closing as CLI arguments are cumbersome for the number of solved cases we are expecting (100+)

PR #321, opened 2023-10-20
update to use new analysis types

MattWellie:
# Fixes

- Production-pipelines now writes report/result entries as metamist analysis type 'aip-report' and 'aip-results'
- This change is not compatible withe current index builder script

## Proposed Changes

  - metamist analysis reporter now writes new entries using those analysis types
  - report finder now detects those analysis types
  - unrelated - seqr lookup dict can be empty/absent without causing failures


cassimons (approved):

MattWellie:
@illusional has updated previous analysis entries to conform to this new typing 🏆

PR #323, opened 2023-10-23
Fix report hunter

MattWellie:
# Fixes

  - The index page (http://popgen.rocks/aip_index) was being published with invalid hyperlinks
  - Reason - the string replacement of `gs://cpg-PROJECT-main-web` with `https://main-web.populationgenomics.org.au/PROJECT` was being done using the `default` project
  - This script runs under the automated account ID `talos`, so the replacement was trying to string replace `gs://cpg-talos-main-web` each time, which was not present in any of the output strings
  - Also tidies up the use of `logging`, so we're not barfing all the metamist results into the hail log every time

## Proposed Changes

  - corrects this so that each analysis object runs a string replacement based on the correct project
  - updates all the CI action versions

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass
- [x]  I've gotten really lazy with these checklist actions

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -23,7 +23,7 @@ jobs:
     - uses: actions/checkout@main
 
     - name: gcloud auth
-      uses: 'google-github-actions/auth@v0'
+      uses: 'google-github-actions/auth@v1'
       with:
         credentials_json: ${{ secrets.GH_IMAGES_DEPLOYER_JSON }}
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,6 +1,5 @@
 name: Docker
 on:
-  pull_request:
   push:
     branches:
       - main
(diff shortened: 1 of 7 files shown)


.github/workflows/docker.yaml
```diff
@@ -1,6 +1,5 @@
 name: Docker
 on:
-  pull_request:
```
jmarshall, line 3: What is the reason for removing this?

helpers/report_hunter.py
```diff
@@ -115,10 +137,9 @@ def main(latest: bool = False):
             exome_output = 'Exome' if 'exome' in output_path else 'Familial'
             singleton_output = 'Singleton' if 'singleton' in output_path else 'Familial'
             try:
-                # pipeline runs don't have display_url
                 report_address = analysis['output'].replace(
-                    get_config()['storage']['default']['web'],
-                    get_config()['storage']['default']['web_url'],
+                    get_config(False)['storage'][cohort]['web'],
+                    get_config(False)['storage'][cohort]['web_url'],
```
jmarshall, line 142: TIL `False` => don't print out the config. Now that is good to know about! :smile:

.github/workflows/docker.yaml
MattWellie: I don't feel super strongly about this change. It was to reduce the overall amount of action runtime. This build is never pushed anywhere, so if we wanted to build and test an image there's a separate `docker_test` wf that will post to `images-dev`. All this one is really doing is proving that a pip install works, which doesn't feel worthwhile to do on each commit.

jmarshall: Works for me. I just like to have the reasons for things spelt out (ideally in the commit messages), for posterity.
(i.e., for future me trying to figure out what's happened…)

MattWellie: aye, that's fair. I was going to add something informative for the removal, but i couldn't think of why this had ever been there in the first place 😓

jmarshall (approved):
LGTM.

All action updates are to the right versions. Some of them were indeed previously erratic — `@main` is a particular problem as the action's author probably doesn't promise that that branch is stable, only tags (usually only major-only tags, i.e. `@vN` rather than `@vN.N.N`) are really recommended.

MattWellie:
> Some of them were indeed previously erratic

I can only imagine this was because I copy-pasted various workflows from all over the CPG and tweaked the steps but not the actions. Good to have nailed this properly

PR #324, opened 2023-10-23
Linter changes

MattWellie:
# Proposed Changes

  - removes flake8 & pylint (and corresponding configs)
  - implements mypy and ruff as replacements
  - remove print-statement-remover pre-commit hook
  - removes all the comments relating to linter suppression


helpers/forbidden_gene_check.py
```diff
@@ -58,9 +58,10 @@ def read_json_from_path(bucket_path: str | Path | None, default: Any = None) ->
     if isinstance(bucket_path, str):
         bucket_path = to_path(bucket_path)
 
-    if bucket_path.exists():
-        with bucket_path.open() as handle:
-            return json.load(handle)
+    if isinstance(bucket_path, Path):
```
illusional, line 61: Do you want an else fail here if you don't pass a string bucket path?

reanalysis/html_builder.py
```diff
@@ -82,10 +82,13 @@ def __init__(self, results: str | dict, panelapp: str, pedigree: Ped):
 
         # Use config to find CPG-to-Seqr ID JSON; allow to fail
         seqr_path = get_cohort_seq_type_conf().get('seqr_lookup')
-        self.seqr = {}
+        self.seqr: dict = {}
 
         if seqr_path:
-            self.seqr = read_json_from_path(seqr_path, default=self.seqr)
+            self.seqr: dict[str, str] = read_json_from_path(  # type: ignore
```
illusional, line 88: I'd add the type annotation (str,str) on the line above, probably helps some typers

reanalysis/utils.py
```diff
@@ -642,7 +641,7 @@ def check_read_depth(self, sample: str) -> list[str]:
         Returns:
             return a flag if this sample has low read depth
         """
-        threshold = get_config()['filter'].get('minimum_depth', 10)
+        threshold = SCRIPT_CONFIG['filter'].get('minimum_depth', 10)
```
illusional, line 644: Not relevant as part of this change, but you could assert filter in the config when you grab it, to save getting to this point without the key

reanalysis/utils.py
```diff
@@ -785,9 +784,10 @@ def gather_gene_dict_from_contig(
         }
     """
 
-    blacklist = []
-    if 'blacklist' in get_config()['filter'].keys():
-        blacklist = read_json_from_path(get_config()['filter']['blacklist'])
+    if 'blacklist' in SCRIPT_CONFIG['filter'].keys():
```
illusional, line 787:
```suggestion
    if 'blacklist' in SCRIPT_CONFIG['filter']:
```

reanalysis/vep_jobs.py
```diff
@@ -288,65 +282,32 @@ def vep_one(
...
     j.image(image_path('vep'))
-    STANDARD.set_resources(j, storage_gb=50, mem_gb=50, ncpu=16)
+
+    # huge resource reduction
+    STANDARD.set_resources(j, storage_gb=50, mem_gb=16, ncpu=4)
 
     if not isinstance(vcf, hb.ResourceFile):
         vcf = b.read_input(str(vcf))
 
-    if out_format == 'vcf':
```
illusional, line 296: Should these changes be in this PR?

illusional (approved):
You have an extra change for VEP that doesn't feel part of this PR - but I'll leave that for you to decide.

I've left a few sporadic changes, but otherwise LGTMichael

reanalysis/vep_jobs.py
MattWellie: ah, this is a bit of bleed from poking around in these scripts. I'll reinstate this, and make all the VEP stuff a separate change

reanalysis/html_builder.py
MattWellie: good spot

helpers/forbidden_gene_check.py
MattWellie: Yup, from the type hinting all supported types are covered, but I've chucked a ValueError if the arg is an invalid type 👍

PR #344, opened 2023-11-28
Introducing models

### LINKED ISSUE #347: Consider a formal schema
The data handling in this application is currently hectic - it's a lot of dictionaries with a lot of post-processing, and the only fixed content it a couple of dataclasses which are still evolving. 

Consider a formally spec'd model to underpin object concepts e.g. the representation of a variant, a MOI-confirmation event, a previously seen variant

https://avro.apache.org/docs/1.11.1/getting-started-python/ might be an example

 This would benefit from an established way of navigating updates to models, such that if the model develops, older versions of the schema can be made compatible

MattWellie:
# Fixes

  - AIP and all it's ancillary scripts have an ill defined data model - everything is just dictionaries, so it's tough to work out what an object is supposed to be at any given time
  - This extends to the serialisation and de-serialisation - this is done by dumping objects to/loading objects from json without any validation. It's entirely possible that the data being persisted doesn't exactly match what code expects as various components evolve over time
  - Closes #347
  - The unit tests all used a various fake versions of 'variants' and 'report entries' with really minimised content. Each of these super minimised entries worked for the specific test case, but wasn't a valid representation of the real objects, making the unit tests ~worthless

## Proposed Changes

  - Introduce Pydantic models for all objects which are currently stored as plain dicts
  - Implement parsing/writing of models to JSON files solely using the pydantic model interface; this includes explicit validation
  - oh my god this is taking so long

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


cassimons (approved):
This is mammoth... but awesome!

PR #348, opened 2024-01-08
Linter changes, fixes scheduled index job

MattWellie:
# Fixes

  - The report hunter process (collect all previous AIP reports into an HTML index page) doesn't work, and hasn't worked in ages
  - It requires a docker image with specific libraries installed, and values in config that weren't previously satisfied in the chrono-triggered runs

## Proposed Changes

  - Runs the whole process in the AIP image (built on top of the driver image as a base, so all dependencies satisfied)
  - No expectation that a `cohorts` section of the config file exists (no need for this check)
  - All functional changes here are contained in `.github/workflows/index_page_builder.yaml` (only run twice a week, not 3 times. Run in the AIP docker image) and report_hunter.py (remove part of the config check), everything else is linting

## Additional

- Following suggestion [here](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1704857573227909), alters the pre-commit hooks, and generates a ton of changes due to the addition of `isort`


.github/workflows/index_page_builder.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch: {}
 
   schedule:
-    - cron: '42 1 * * 1,3,5'
+    - cron: '42 1 * * 2,4'
```
illusional, line 7: sShould put a comment here about what this functionally means

illusional (approved):
This is your repo, so I'll let you own the title, but this to me is a Linting PR + deploy docker update and I think the PR title should reflect that to make it easier in the future to make it easier to ignore the commit in blames (and be more descriptive in git logs).

MattWellie:
> This to me is a Linting PR + deploy docker update and I think the PR title should reflect that

Yeah this was rough timing - it was an isolated workflow fix, then you mentioned isort on Slack. I didn't think it would change quite so many files... Title is now pretty misleading, so i'll sort that out

Issue #366, opened 2024-03-14
CTX SV variants break parser


MattWellie:
# CTX

```SVTYPE=CTX. BreakDancer-style notation used by TIGRA to indicate an inter-chromosomal breakpoint```

These variants don't have an SVLEN attribute, that would make no sense. Parser needs to handle breakpoints like this.

https://batch.hail.populationgenomics.org.au/batches/435991

MattWellie:
Solved

PR #367, opened 2024-03-14
Allow for interchromosomal split variants

MattWellie:
# Fixes

  - #366 

## Proposed Changes

  - Inter-chromosomal variants don't have a length. Instead, we set the 'SVLEN' to be 'chrX:12345' (the other side of the breakpoint)

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/minimise_output_for_seqr.py
+++ b/reanalysis/minimise_output_for_seqr.py
@@ -15,10 +15,10 @@ Also produce a second version of the same, limited to phenotype-matches
 """
 
 import json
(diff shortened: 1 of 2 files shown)


cassimons (approved):
So will this make genie breakpoints in scope?

MattWellie:
> So will this make genie breakpoints in scope?

Yup! So long as the breakpoint causes an LoF annotation we're already picking them up... Unless this also fails in some new weird way

PR #371, opened 2024-03-19
Category 6 & Support Changes

### LINKED ISSUE #370: Proposal: AlphaMissense Ascendant
Re-work the categories to adjust to the facts:
- Support (determined through _in silico_ predictors) largely overlaps with Cat6 (AlphaMissense prediction)
- c6/AM is noisy for cohorts without family structure to better filter for MOI matches

## Proposed changes

1. Replace the CADD/REVEL involvement in Cat. 2 with AM pathogenicity
2. Remove the CategorySupport logic. The code placeholders to trigger 'support' logic will remain, but we just won't exercise those for now.
3. Add back-filtering logic to remove variants after MOI application where they are Cat6 only, and the variant-gene has not been phenotype-matched to this participant

## Future

- Turn off Category 2? This will be done in config for now, but after evaluation we may drop this category. The first seen/latest report takes care of what we wanted to accomplish with this logic, and is less volatile (results coming & going for each run)

MattWellie:
# Fixes

  - Closes #370 

## Proposed Changes

  - Removes the Support category, and the in silico tools we were using (MutationTaster, Polyphen, Sift, CADD, REVEL)
  - AlphaMissense takes a more substantial role (e.g. as a supporting element in Cat 2)
  - AlphaMissense scoring threshold doesn't change, but by default it becomes present on the naughty step `these categories require a phenotype match` list
  - If after MOI testing a variant is:
       - unsupported in a comp-het
       - isn't in a gene on a phenotype- or cohort-forced panel
       - all assigned categories are on the naughty step
  - we remove it from the analysis

## Testing

needs testing, but first I need to downgrade my Java installation as JRE 21 doesn't support Spark, so Hail breaks

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

Pretty substantial other change - the Dockerfile base is changed from the CPG AR driver image to a hail-genetics base. Shaves off almost 4GB from the build. Should be fine as we don't invoke gcloud/gsutil directly


cassimons (approved):
Looks good to me.

PR #378, opened 2024-04-11
Change HPO term storage

MattWellie:
# Fixes

  - Not sure if this is required, setting as a draft
  - I used to store HPO terms as a list of Strings. Recently this changed to `HPO - Text Description`. That packaging is going to make it harder to pull out just the HPO terms 

## Proposed Changes

  - Takes the list of compound Strings and changes it to a dictionary of `{'HPO': 'Description'}`
  - Later it should be pretty easy to pull out just the HPO terms if required for phenotype matching

## Checklist

- [x] Linting checks pass

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,56 +1,39 @@
+default_language_version:
+  python: python3.10
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
     hooks:
       - id: check-yaml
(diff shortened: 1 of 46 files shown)


cassimons:
Do you want this in now? Seems like a good idea to me.

MattWellie:
> Do you want this in now? Seems like a good idea to me.

Can do! Do your pheno-matching changes rely on this? The conflicts are wild, will try and clean that up

MattWellie:
Crazy diff, get in the bin

PR #382, opened 2024-04-17
Environment Lock

nevoodoo:
## Description
In a bid to introduce better security permissions, we have replaced the use of SA keys to workload identity providers. This change will also implement the use of environments to ensure deployments are going through the right channels.

## Changes
- [x] Updated the workflow to use different WIP for dev
- [x] Added `environment` attribute to the `docker` workflow
- [x] Split the `docker` workflow into two jobs, one for prod and one for dev


MattWellie (approved):
LGTM

PR #383, opened 2024-04-17
show_default fix

MattWellie:
# Fixes

  - `show_default` is not a valid argument to argparse - CoPilot led me astray. Looks like `show_default` is a valid argument for the Lua version of the library, but not for Python, so this was causing startup failures.
  - A while back I removed some checkpoints from the Hail process - the recent runtimes are WAY higher than I would expect, so reinstating some checkpoints to see if that alleviates the issue.

## Proposed Changes

  - Remove bad argument from argparse
  - adds checkpoints back in to match a previous state ([here](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/2fbb058d728bf12d82c6c4b14c7c118a64ab8dd4/reanalysis/hail_filter_and_label.py))

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/reanalysis/hail_filter_and_label.py
+++ b/reanalysis/hail_filter_and_label.py
@@ -916,13 +916,7 @@ def subselect_mt_to_pedigree(mt: hl.MatrixTable, pedigree: str) -> hl.MatrixTabl
     return mt
 
 
-def main(
-    mt_path: str,
-    panelapp_path: str,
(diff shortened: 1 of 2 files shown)


cassimons:
Co-pilot did you dirty!

cassimons (approved):

PR #390, opened 2024-04-25
Add gene level hpo matching to main workflow.

cassimons:
## Proposed Changes

Attempts to add genlevel phenotype matching to main workflow. This is currently broken, but I thought it might help discussion. 

Uses Semsimian package to compare patient phenotypes with phenotypes associated with gene. If one or more terms are above the similarity threshold then it is considered a phenotype match.

Currently reports the result as part of the panels object. This "fits" in terms of being functionally similar, but we may want to think about renaming this.

Questions:
- Is this the right place to put this, or should it go somewhere else?
- where do we handle regular downolad up to date versions of phenotype and hpo db files? Needs to ne done each month-ish.
- genes_to_phenotype file has gene symbol and NCBI gene ID. How/where do we map to ENSGIDs

Todo:
- [ ] Add  Semsimian package as a dependency
- [ ] Download up to date versions of phenotype and hpo db files. 
- [ ] Localise phenotype and hpo db files and pass paths
- [ ] Add min similarity score to config

### DIFF
--- a/reanalysis/models.py
+++ b/reanalysis/models.py
@@ -297,6 +297,7 @@ class ReportPanel(BaseModel):
 
     forced: set[str] = Field(default_factory=set)
     matched: set[str] = Field(default_factory=set)
+    gene_level_phenotype_matches: set[str] = Field(default_factory=set)
 
 
 class ReportVariant(BaseModel):
--- a/reanalysis/validate_categories.py
+++ b/reanalysis/validate_categories.py
@@ -19,6 +19,7 @@ from collections import defaultdict
 
 from cyvcf2 import VCFReader
 from peddy.peddy import Ped
+from semsimian import Semsimian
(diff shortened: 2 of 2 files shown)


MattWellie:
I wasn't sure in general how this would slot in - this could be a good fit, adding in the priority annotation during result generation.

Alternatively we add a separate polishing step for the results, where we could feed in semsimian, exomiser, and any other 'preferred genes for cohort/participant' input. I wouldn't want to drop all those alternatives into this stage as it'll become quite messy. That would separate out the 4 phases into distinct runtimes
- Variant labelling 
- MOI testing
- Phenotype polishing (maybe migrating the panel/participant and panel/cohort labels into this stage too)
- Report generation

MattWellie:
Closed after #426

PR #392, opened 2024-04-26
Adds versions to models

MattWellie:
# Fixes

  - The Models used are defined in code, but on file as JSON. We currently have no way of determining whether the current code is correct to parse the model JSON

## Proposed Changes

  - adds a separate model version ID in the codebase (in models.py)
  - adds this model version ID to the JSON objects written to file as a top-level attribute
  - Future changes to the models can cause a change in version number, so reading an object from file becomes a multi-step process:

1. read the model version attr from the JSON data
2. if current != version in file, read as JSON, pass through an adapter method, then parse as a current object (e.g. adding in missing mandatory fields, or changing a string to a list of strings)
4. if current == version in file, parse JSON directly

This will be a starting point, and any breaking future changes to the models will have to come with a method to preserve backwards compatibility

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


cassimons (approved):
I like

PR #395, opened 2024-05-01
Model liftover methods

MattWellie:
# Fixes

  - #393 introduced a change to the data model, but AIP had no way to automatically update data written in a prior version of the data models
  - That wasn't a breaking change, but we'll probably have breaking changes which pydantic will cry about in the future

## Proposed Changes

  - Each top-level model now has a `version` attribute (was `model_version`, but this was a protected name in pydantic...). These top levels are chosen so that each persisted-to-file JSON dict has version as a top level key.
  - A liftover method registry is created, so that each transition between versions is accounted for. Each transition method should take data on one version of the model, and export data valid on the next model. If the versions are multiple releases apart, this chain can include multiple methods
  - If there is no liftover required for a model, no update method is registered
  - In future, when the model version number changes, any liftover methods need to be created and tests added to prove function

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


reanalysis/utils.py
```diff
@@ -182,7 +183,7 @@ def get_cohort_seq_type_conf(dataset: str | None = None):
     """
     global COHORT_SEQ_CONFIG
     if COHORT_SEQ_CONFIG is None:
-        dataset = config_retrieve(['workflow', 'dataset'], dataset)
+        dataset = dataset or config_retrieve(['workflow', 'dataset'])
```
MattWellie, line 186: I'm sneaking this unrelated change in because I broke AIP in the pipeline :D

.github/workflows/clinvar_runner.yaml
```diff
@@ -32,5 +32,5 @@ jobs:
...
-              -d '{"output": "generate_clinvar_${{ steps.date.outputs.date }}", "dataset": "talos", "accessLevel": "full", "repo": "automated-interpretation-pipeline", "commit": "${{ github.sha }}", "cwd": "reanalysis", "script": ["./clinvar_runner.py"], "description": "Generate Latest Clinvar Summaries", "image": "australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_aip:3.3.0", "config": {"workflow": {"sequencing_type": "genome"}, "cohorts": {"talos": {"clinvar_filter": ["victorian clinical genetics services, murdoch childrens research institute"]}}}, "wait": false}' \
+              -d '{"output": "generate_clinvar_${{ steps.date.outputs.date }}", "dataset": "talos", "accessLevel": "full", "repo": "automated-interpretation-pipeline", "commit": "${{ github.sha }}", "cwd": "reanalysis", "script": ["./clinvar_runner.py"], "description": "Generate Latest Clinvar Summaries", "image": "australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_aip:3.3.1", "config": {"workflow": {"sequencing_type": "genome"}, "cohorts": {"talos": {"clinvar_filter": ["victorian clinical genetics services, murdoch childrens research institute"]}}}, "wait": false}' \
```
EddieLF, line 35: Nitpick but could the versioning here for the aip image could be clearer at the top, like the versioning in the docker.yaml? It's kinda buried in a long line here

EddieLF (approved):
A++ This was a pleasure to review

.github/workflows/clinvar_runner.yaml
MattWellie: Updated! One of the github CI files was already using this, I've expanded that so that the diff will be much clearer

PR #398, opened 2024-05-08
Truncate and source column

MattWellie:
# Fixes

  - Super long indels are ruining the report formatting

## Proposed Changes

  - Inserts some logic to detect ref or alt > 10bp, and changes the long LOLOLOLOLOLOLOLOLOLOLOLOLOLOLOL>LOL to `ins 34bp`
  - Adds a liftover method for ReportData
  - Adds a Source column (SmallVariant or StructuralVariant) to make the two variant types clear (the `INS Nbp` display is already in use for SVs)
  - Moves some logic out of the Jinja template (probably doable in Jinja instead, but looks messy)


reanalysis/html_builder.py
```diff
@@ -359,21 +359,44 @@ class Variant:
...
+        - SVs always need to be presented differently
+           - e.g INS 4079bp
+        """
+        if isinstance(self.var_data, SmallVariant):
+            if len(self.ref) > 10 or len(self.alt) > 10:
+                ref_len = len(self.ref)
+                alt_len = len(self.alt)
+                if ref_len > alt_len:
+                    return f'del {ref_len - alt_len}bp'
+                return f'ins {alt_len - ref_len}bp'
```
EddieLF, line 376: If the ref and alt are the same length (and >10bp) will we see "ins 0bp" in this field? Is that fine?

EddieLF (approved):
Not too familiar with Jinja, but the python changes look good 🐍

reanalysis/html_builder.py
MattWellie: I'll give that a little tweak - I don't think I've ever seen a variant with both huge ref and alt alleles, but it's possible

MattWellie: this scenario will result in `complex delins nbp`

PR #402, opened 2024-05-13
Never reuse Hail checkpoints

MattWellie:
# Fixes

  - Code so far has re-used checkpoints by default. That has lead to numerous dodgy results where input files/panels were fixed, but the re-run picked up the previous run's content

## Proposed Changes

  - Re-running is now not the default
  - Can be overridden in config
  -

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


cassimons (approved):
Kill them with fire.

PR #407, opened 2024-05-31
Steps towards making AIP more generic

MattWellie:
# Proposed Changes

Frankly, a ridiculous number of changes

### Chunky Changes

- [cpg_generate_pheno_ped.py](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/make_it_more_generic/talos/cpg_generate_pheno_ped.py): CPG implmentation of a new pre-pipeline step generating an extended PED file from a metadata service
- [vep_vcf_to_mt.py](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/make_it_more_generic/talos/vep_vcf_to_mt.py): new script to translate a VEP'd VCF to a MT

### Various other
  - removes metamist file/analysis registration completely out of this codebase (we handle that at a pipeline level, this code is specific to CPG and no longer used)
  - removes metamist connection from the panel matching stage
  - creates a new script to connect to metamist and parse all required details into an extended PED file format. This is intended to be the only metamist-specific entrypoint, which other users could swap out for their own implementation
  - drops reliance on Peddy, instead using `peds`, parsing into a private model where required
  - MOI testing is done as independent trios, not as a whole family (i.e. for each affected participant, variant(s) will be accepted if the inheritance model works within the proband-parents trio. An affected/unaffected sibling/grandparent is not considered for the MOI to hold)
  - NEW NAME

## Next steps

 - SV/CNV VCF -> MT adapter
 
## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


talos/local_runner.py
```diff
@@ -0,0 +1,309 @@
+#!/usr/bin/env python3
+
+
+"""
+Simulated local run of Talos - Hail Query using a local runtime
+Currently applies only to SV data as a minimal test case
```
cassimons, line 6: Is this still true?

test/input/seqr_tags.tsv
```diff
@@ -1,3 +1,3 @@
 chrom	pos	ref	alt	gene	worst_consequence	1kg_freq	exac_freq	gnomad_genomes_freq	gnomad_exomes_freq	topmed_freq	cadd	revel	eigen	splice_ai	polyphen	sift	muttaster	fathmm	rsid	hgvsc	hgvsp	clinvar_clinical_significance	clinvar_gold_stars	filter	family	tags	classification	notes	sample_1	num_alt_alleles_1	gq_1	ab_1	sample_2	num_alt_alleles_2	gq_2	ab_2	sample_3	num_alt_alleles_3	gq_3	ab_3
-17	10697288	G	A	SCO1	missense_variant	0	0	0	0	0	15.34	0.195	0.602	0	benign					ENST00000255390.9:c.220C>T	ENSP00000255390.5:p.Pro74Ser				fam001108	Review|VUS|Known gene for phenotype|AIP training: Possible			PROBAND	2	99	1	FATHER	1	99	0.5147058823529410	MOTHER	1	99	0.42857142857142900
+17	10697288	G	A	SCO1	missense_variant	0	0	0	0	0	15.34	0.195	0.602	0	benign					ENST00000255390.9:c.220C>T	ENSP00000255390.5:p.Pro74Ser				fam001108	Review|VUS|Known gene for phenotype|Talos training: Possible			PROBAND	2	99	1	FATHER	1	99	0.5147058823529410	MOTHER	1	99	0.42857142857142900
```
cassimons, line 2: Dou you wan this changed?

cassimons (approved):
This looks awesome!

talos/local_runner.py
MattWellie: This is definitely not true, but I'll remove this script before merging. It was just a test case for the AIP-in a bottle use case

test/input/seqr_tags.tsv
MattWellie: Nope! That's a mistake

PR #413, opened 2024-06-25
Add entrypoints

MattWellie:
# Fixes

  - the separate scripts that make up the Talos workflow are being referenced by paths to scripts inside the container

## Proposed Changes

  - adds a console_script entrypoint for each separate script
  - I've split out the entrypoint within those scripts to have a cli-accessible main method (where CLI arguments are parsed), and a code-accessible main method (so that I can add a simpler single-step runner for easier use)
  - this PR is almost entirely moving things around, not changing behaviour
  - This will require a PR in prod-pipes

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


talos/hail_filter_sv.py
```diff
@@ -136,20 +149,30 @@ def main(mt_path: str, panelapp_path: str, pedigree: str, vcf_out: str):
         pedigree ():
         vcf_out (str): where to write VCF out
     """
+    get_logger(__file__).info(
```
illusional, line 152: Should define this somewhere central to reuse, given everything is already a script

talos/vep_vcf_to_mt.py
```diff
@@ -28,6 +24,46 @@
     'string': {'type': hl.str, 'default': ''},
     'float': {'type': hl.float64, 'default': 0.0},
 }
+RELEVANT_FIELDS = relevant_fields = [
```
illusional, line 27: Is this double assign intentional?

illusional (commented):
I quickly scanned, and broadly looks fine - but not explicitly approving because Cas is listed on here, and there are fairly substantial changes outside just CLI + entrypoints (like checkpointing changes)

cassimons (approved):

MattWellie:
> like checkpointing changes

This was bundled in just to reduce the number of CLI args, a slight behaviour change. We're doing everything in a container and throwing the checkpoint away, so removing the need to pass a specific path. This behaviour might return when we need to support other environments where they might want to recover from a partially-complete run using a persisted checkpoint

talos/vep_vcf_to_mt.py
MattWellie: nope! removed

PR #419, opened 2024-07-16
Use specific checkpoint location

MattWellie:
# Fixes

  - To complete Hail-Query runs in good time, we need to write a checkpoint
  - Hail reliably writes that to the OS partition on the device, with very little spare room
  - This kills the VM and the run

## Proposed Changes

  - Checkpoint path becomes a String, or None
  - If it's a String, that's where the checkpoint goes
  - If it's none, we don't write a checkpoint

This is re-implementation of some old behaviour. Previously I wrote 2 checkpoints, the first could be controlled in this way, the second could not. No explanation from Hail Team. 

With one checkpoint we should be good. 

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


cassimons (approved):
> Previously I wrote 2 checkpoints, the first could be controlled in this way, the second could not. No explanation from Hail 

This is so odd...

PR #420, opened 2024-07-17
Correction to report splitting

### LINKED ISSUE #421: Runtime Failures on CNV processing
The new PED generation script has not been sequencing-type specific

When we gather all the relevant sample IDs, if a participant has an exome and a genome, the ped will only retain one.

This has lead to issues where the SV MTs are being processed and no overlap is found between the pedigree and MT/VCF

This runtime bug masks the deeper issue of 'some samples are arbitrarily being excluded from the analysis because they have a genome sample, and this is an exome run', or vice versa

Suggestion - limit the PED generation to exome or genome as appropriate. Delete the latest runs and regenerate

### LINKED ISSUE #422: Failing to write report subsets
I've used Pathlib.Path().parent to get to a parent directory of a filepath.

This is because at runtime this can create an unpredictable number of paths, due to splitting on content.

- The use of Pathlib.Path on a `gs://...` address removes the double `//`
- cloudpathlib interprets a `gs:/...` prefix to be a local path, and can't open the requested file path as a local file
- Result was that for the cohorts we now split (due to being... massive) we now write the main and latest report, then the report generation script breaks, so we don't get the mini-reports

MattWellie:
# Fixes
  - Closes #421 
  - Closes #422
  - When generating multiple reports for a cohort (splitting by year, or an arbitrary split), I was deriving the output folder to write additional reports to from `Path(<main_output_file>).parent`
  - The use of Pathlib.Path on a `gs://...` address removes the double `//`
  - cloudpathlib interprets a `gs:/...` prefix to be a local path, and can't open the requested file path
  - Result was that for the cohorts we now split (due to being... massive) we now write the main and latest report, then the report generation script breaks, so we don't get the mini-reports
 
## Second issue - #421 

In the Hail analysis of SV data, the process quits if there's no overlap between the PED and VCF

This has caused recurrent failures where a participant has both Exome and Genome results - only the exome data is being retained in the pedigree, so there appears to be no overlap with the exome-only CNV data

## Proposed Changes

  - Stop using Pathlib.parent to get the directory name
  - Instead just get the file name, and remove from the full path to get the directory
  - use plain string manipulation to build the path after that

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


MattWellie:
also took this opportunity to delete more linting exclusions

src/talos/GeneratePED.py
```diff
@@ -22,10 +22,10 @@
 HPO_RE = re.compile(r'HP:[0-9]+')
 PARTICIPANT_QUERY = gql(
     """
-query MyQuery($project: String!) {
+query MyQuery($project: String!, $sequencing_type: String!) {
   project(name: $project) {
     pedigree
-    sequencingGroups {
+    sequencingGroups(type:  {eq: $sequencing_type}) {
```
EddieLF, line 28: You could consider adding a filter for sequencing group technology as well, like `technology: {eq: "short-read"}`.

I'm just noting that below, this query is parsed and each sequencing group is keyed by the participant external ID. In some cases, we can have participants with multiple sequencing groups of the same sequencing type but  with differing technologies (or platforms). 

If you were parsing a project that had a participant with both a "long-read" and "short-read" sequencing group of type = "genome", then you will lose one of these sequencing groups as you construct the `ext_to_int` dict.

MattWellie: Nice, hadn't really considered that angle

EddieLF (approved):
Nice 🌱

PR #431, opened 2024-09-03
expose denovo function config

cassimons:
## Proposed Changes

  - Exposes aditional config options ued by hail denovo function


MattWellie:
Now handled in main

PR #434, opened 2024-09-26
Feature 433: Remove HPO tree limit

### LINKED ISSUE #433: HPO tree traversal should be unbounded
When 'fuzzy matching' between the HPO terms in a family, and HPO terms annotated on a family, we allow for a tolerance of 3 tiers separation in the HPO hierarchy between the two terms. This is arbitrary, and has not been extensively evaluated.

We should consider this becoming an unlimited scan to the root of the HPO tree, and see how that affects results.

Value [here](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/src/talos/GeneratePanelData.py#L23)

Used [here](https://github.com/populationgenomics/automated-interpretation-pipeline/blob/main/src/talos/GeneratePanelData.py#L149)

MattWellie:
# Fixes

Closes #433 

  - When fuzzy-matching between participant and panel HPO terms we search 3 levels apart in the HPO ontology tree. If no matches are found within that 3-level search, quit searching and move on to the next
  - This 3-level search limit was never based in anything, and nothing has been done to validate that 3 is a good number
  - Limiting the search to 3 tiers also implicitly requires a degree of coordination between recruiting sites and PanelApp curators, which we do not have in place. There's nothing to suggest that the data we process from participant cohorts and PanelApp have tuned matches to within 3 levels of HPO hierarchy, so it's likely we've been missing out on valid matches by cutting our searches short.

## Proposed Changes

  - Removes the max-level cap of 3 completely. The search now becomes 'unbounded', all the way from query term to HPO root

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/src/talos/GeneratePanelData.py
+++ b/src/talos/GeneratePanelData.py
@@ -20,7 +20,6 @@ from talos.models import ParticipantHPOPanels, PhenoPacketHpo, PhenotypeMatchedP
 from talos.static_values import get_logger
 
 HPO_RE = re.compile(r'HP:[0-9]+')
-MAX_DEPTH = 3
 
 PANELAPP_HARD_CODED_DEFAULT = 'https://panelapp.agha.umccr.org/api/v1/panels'
 try:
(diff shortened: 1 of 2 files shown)


cassimons (approved):
All good other than that failing test...

PR #437, opened 2024-10-01
Remove cat2, change notion of new

### LINKED ISSUE #436: Changes to Cat2/Recent
From #435 

```
Now that we have dropped cat2, we need to find a way to highlight any variant in "recent" new genes. My straw man for how this should work:

- Have a configurable time delta of X months. Any gene that is "new" to panelApp in the past X months should get a "New" label

- It would be great to distinguish between "new" to the mendeliome and "new" to and new to a family-specific panel. Can we make this distinction easily?

- In the report we indicate new genes with a gold ⭐ next to the Gene symbol OR a blue ⭐ if it is just been added to a panel
```

Just to confirm up-front - implementing this would mean dropping Category 2 (and its associated more lenient logic) completely? 

We'd instead be in a situation where variants in any category can be flagged as 'new', but being new itself doesn't feed into the category analysis.

---

This all sounds doable using the Activity endpoint in PanelApp. I included something previously where I used the panel activity endpoint to poll for a specific panel version based on dates, but that was... janky, and differences between two versions of the same panel were kinda unsatisfying. 

Instead I can pull the Panel data as normal, then the activity log for the panel. That contains a date stamp for the event where each gene is marked as Green, or 'ready', e.g.

```
{
    "created": "2024-09-11T07:38:43.562302Z",
    "panel_name": "Mendeliome",
    "panel_id": 137,
    "panel_version": "1.2011",
    "user_name": "Zornitza Stark",
    "item_type": "entity",
    "text": "Marked gene: RNU2-2P as ready",
    "entity_name": "RNU2-2P",
    "entity_type": "gene"
  },
  {
    "created": "2024-09-11T07:38:43.547129Z",
    "panel_name": "Mendeliome",
    "panel_id": 137,
    "panel_version": "1.2011",
    "user_name": "Zornitza Stark",
    "item_type": "entity",
    "text": "Gene: rnu2-2p has been classified as Green List (High Evidence).",
    "entity_name": "RNU2-2P",
    "entity_type": "gene"
  },
```

It would be a case of parsing these log entries for all green genes, identifying the point at which each gene was marked as green, and seeing if that falls in the pre-determined time delta to classify it as 'new'.

---

Time Delta - this would be in config, but 2 months sounds like a decent window. Maybe 2.5? We typically run reports every month, so that would be enough time that an analyst could miss one report cycle without completely missing that a gene was recently added. The .5 buffer would mean that there's some leeway in what specific part of the month the reports are run.

This wouldn't interact specifically with the 'first seen' stamp. A variant could appear as 'new' for multiple reports in a row, but the first date seen would be a fixed value. Analysts are always free to filter the report using `> DATE` syntax.

---

New in the Mendeliome vs. phenotype matched panel would be relatively simple - the notion of 'new' I've been storing is a list of panel IDs, rather than a Boolean. That's already in use when deciding if Cat. 2 applies to each individual in the analysis, so I'd keep that behaviour, and render it in the report with either a mendeliome or a feature panel star (or both).

---

This report is going to need a figure legend for the various icons and tooltips 😆

MattWellie:
# Fixes

  - Category 2 has always been a weird one. It's not about the variant annotations themselves, but based entirely around what happens in one analysis vs. the previous analysis
  - Cat 2 also adds leniency to the process - if a gene is new in a panel we make it easier for inconsequential results to reach the report
  - In combination this means that a gene that's new in the analysis is really likely to reach the report once, then drops back to a standard baseline
  - Closes #436

## Proposed Changes

  - Remove Cat 2. It's gone. Get outta here
  - Add a new way of looking for new variants - when we pull the PanelApp contents for a panel, we also pull the activity log. If a gene was added to the panel within the last X months (configurable in config), it will be highlighted in the report with a Star

## Also Changes

- There's a few model changes to how and where bits of panel-related data is being stored
- The 'PhenotypeMatch' (Semsimian) is not a separate column, this has also been moved over as an accent on the gene
- Appearance of Star accents:
Screenshot 2024-10-01 at 6 07 33 PM

Screenshot 2024-10-01 at 6 07 51 PM

The Green/Blue man indicator is the same: present in a phenotype matched panel / a variant-level match to the family HPOs

The stars are new: Gold for new in the Mendeliome, Blue for new in one or more matched panels (tooltip shows the panel names and numbers)

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


cassimons (approved):
Amazing thank you! May Z forgive us 😉

PR #440, opened 2024-10-08
A couple of presentation corrections

MattWellie:
# Fixes

  - forced_matches was a set of strings, not a dict - I missed that
  - Analysts have requested that the Variant level matching tooltip is updated to also be a person. Little Man = match priority, Star = gene is new
  - There was an issue whereby some PanelApp entries were dumped onto panels with only semi-structured activity log entries. This has caused some old entries to appear new

## Proposed Changes

  - update the report icons, and edit the colours and tooltips
  - expand PanelApp activity search to catch earlier unstructured entries
  - Correct how forced_matches is being processed


cassimons (approved):
Looks good to me.

Not sure if you want to put it in this or separate, but I think we need to enable the ability to filter on a phenotype match. I assume that means moving the little men to their own column called "Phenotype match" then adding some text (eg "Y" or "Yes") that can be used to filter on. I am tempted to have the filter set by default.

MattWellie:
We just dropped the Phenotype Match column, which was asked for here https://github.com/populationgenomics/automated-interpretation-pipeline/issues/435

I like where the phenotype match icon is at the moment, so we could add a separate column at the far edge which just had a simplified Y/N for filtering on. I'm not sure if the HTML columns we're using at the moment will allow filter on by default? If they do that's simple. 

At the moment we produce 2 final reports, one restricted down to just phenotype matches, the other with everything. We probably don't want to use the match-only here as it would limit searching

PR #441, opened 2024-10-08
Report polishing

MattWellie:
# Fixes

  - Column header for ClinVar is kinda misleading, we're using ClinvArbitration (our own reinterpretation)
  - We've not got a way to filter down to only phenotype matched genes
  - There's now 5 separate icons used to show different pieces of information - these are not explained, other than in a hover-over tooltip

## Proposed Changes

  - Adds a simple phenotype match column, containing Y/N. Y if the variant is a cohort/panel/variant level match, else N
  - Column header corrected to say ClinvArtbitration
  - Table legend added which explains each of the main elements, and links out to the ClinvArbitration repository


cassimons (approved):
Looks good. I agree the link in the header is a bit janky. Perhaps we leave that for the mo then revisit with a docs page or some other way of explaining what is what.

PR #450, opened 2024-10-22
Trust SV qualities

MattWellie:
# Fixes

  - We're missing some CNV calls due to whole sites being failed due to a failing Q score at any one component participant

## Proposed Changes

  - Scrap the Q score filter when processing SVs
  - ... that's it
  - CyVCF2 doesn't have a way to check out the Q scores on the VCFs coming out of SVAnnotate via Hail, so it's non-trivial to flag some of these Q failures later.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,7 @@ protobuf>=3.20
 pydantic>=2.5.2
 pyspark>=3.5.1
 python-dateutil>=2
-requests>=2.31.0
+requests>=2.32.0
 semsimian>=0.2.15
 tabulate>=0.8.9
 toml==0.10.2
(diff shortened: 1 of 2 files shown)


cassimons:
Can/should/do we just do variant-level GQ filtering on SVs? (default > 20) ?

cassimons (approved):

MattWellie:
> Can/should/do we just do variant-level GQ filtering on SVs? (default > 20) ?

yes, I think so :D I'll do some poking to see if the site QUAL is intelligent, or taken from the first VCF in the merge. I'll get this in for now so we can see whether a completely unfiltered approach is viable

PR #453, opened 2024-10-30
Bugfixes, and one fast report

MattWellie:
# Fixes

Miah has spotted a couple of bugs
- the arguments to the PED -> PhenoPacket converter need correction
- the HTML generator script doesn't tolerate an output path outside of the current directory

## Proposed Changes

  - correction to the ped conversion arguments
  - use os.makedirs based on the target location, not the current location

## Report Fixes

- for the CPG pipeline, we need to create the main document: The CPG pipeline will only register outputs in metamist if the exact file has been created, so we always make that, then optionally do splitting.
- the jinja template engine is now started up with `trim_blocks` and `lstrip_blocks` - these arguments stop jinja from inserting empty lines in the HTMLs we produce. These account for over half the file length in lines, but their removal didn't speed anything up
- I've also fixed the clinvar hyperlink - I was creating a hyperlink with no content if the allele doesn't have a clinvar entry. This removal also didn't meaningfully speed things up, but it's silly.
- This includes a complete GPT-4 inspired revolution on the report - now it's paginated, with a spectacular performance upgrade.
- Sample/Family rows are now collapsible using [this functionality](https://live.datatables.net/cecosoru/1/edit)

## still required

- now that we don't need to split report, we can try to reconcile this new jquery with the original child table functionality

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


MattWellie:
something like this would be useful to combine the child-row details with the main view, but DataTables seems to be less straightforward than it was using TableSorter
e.g. https://datatables.net/dev/accessibility/DataTables_1_10/examples/api/row_details.html ?

src/talos/templates/index.html.jinja
```diff
@@ -72,42 +101,61 @@
...
+                // Uses the 'row group' plugin
+                dataSrc: 0,
+                startRender: function (rows, group) {
+                    var collapsed = !!collapsedGroups[group];
+
+                    rows.nodes().each(function (r) {
+                        r.style.display = collapsed ? 'none' : '';
+                    });
+
+                    // Add category name to the <tr>. NOTE: Hardcoded colspan
```
jeremiahwander, line 125: If it stresses you out to not have the row group headers run the entire width of the table consider the following

`var colspan = $('#variant-table thead tr th').length;`

Followed with a modification to the generated HTML

`.append('<td colspan="' + colspan + '">' + group + ' (' + rows.count() + ')</td>')`

cassimons (approved):
This is amazing. The only issue I can see is that the tooltips popups are broken, but I think that just need adding back into the js?

MattWellie:
Ah yeah, I copied that back in locally but couldn't solve child rows so didn't push anything

src/talos/templates/index.html.jinja
MattWellie: I was taking that code block (including the NOTE) from a JS example. I've just traded this out for 

`<td colspan="100%">`

... I don't know if there's a reason not to do that, seems like the original example I copied would have gone that way

PR #454, opened 2024-11-05
Strip out MOI-related HPO terms

MattWellie:
# Fixes

  - Some of our cohorts have been tagged with MOI-related terms, e.g. "Autosomal Dominant"
  - This MOI-related terminology is being recovered as matches to genes, where the gene is not related to the participant's condition, but is related to the MOI of the gene itself

## Proposed Changes

  - During generation of the phenopackets file we groom our any HPO terms from consideration which are descendants of HP:...5, "Mode of Inheritance"
  - This only needs to be done once, early on. Each use of HPO terms later on stems from this original file
  - If no HPO file is supplied, we don't remove terms, but we also don't need to - no HPO-matching will take place


cassimons (approved):
Great thanks.  Given this is an active decision we are making, would you mind adding a bit more detail on the reasoning to the comment in line 72?

PR #455, opened 2024-11-06
Nextflow runner

MattWellie:
# Fixes

  - Talos lacks a demonstration of what an end-to-end workflow run looks like, and a convenient way of getting there
  - Increasingly important if this is shopped around as an analysis tool, other centres need a way to kick the tires

## Proposed Changes

  - Installs Nextflow into the Docker image
  - Creates a full E2E run using spoofed data
  - demonstrates a run of VCF -> MT -> Categorisation -> Results -> HTML
  - This framework in general should be useful as an integration test - that will require more fleshing out in terms of validating a specific results set etc. 

Stuff this still needs: 

1. building a container, then operating commands inside it
    - DONE, so far this is a demo of how to run the whole pipeline end to end where each stage is containerised, which requires nextflow to be installed to run the demo. It could be redrafted to run a container, with a full nextflow workflow inside it, in which case a user wouldn't need to use nextflow. I'm not sure that's the best model - the computational resource demands of the different stages are wildly different, so probably best to keep all stages as separate tasks, and we can whip up some suggested templates for each.
3. checking the outputs of each stage against an expected result
4. working out how best to either reduce the number of required inputs, or reduce overhead getting it started


talos.nf
```diff
@@ -0,0 +1,235 @@
...
+
+workflow {
+    // existence of these files is necessary for starting the workflow
+    // we open them as a channel, and pass the channel through to the method
+    pedigree_channel = Channel.fromPath(params.pedigree)
+    hpo_file_channel = Channel.fromPath(params.hpo)
+
+    // turn the VCF into a MatrixTable
+    input_vcf = Channel.fromPath(params.annotated_vcf)
+    input_vcf_idx = Channel.fromPath("${params.annotated_vcf}.tbi")
```
Alexander-Stuckey, line 205: You can do a map to make a vcf and index in the same variable (I don't remember the exact syntax).

```
input_vcf_and_index = Channel.fromPath(params.annotated_vcf)
map{ it -> [file(it), file("${it}.tbi")]}
```

talos.nf
```diff
@@ -0,0 +1,235 @@
+
```
Alexander-Stuckey, line 1: The general format for writing a DSL2 workflow in Nextflow now is to have each process code in a separate file under a `modules` directory, and then import the needed modules at runtime.
e.g. https://github.com/nf-core/smrnaseq/blob/72c4c4cf83e1f534319493f9de8344854f59736f/subworkflows/nf-core/bam_sort_stats_samtools/main.nf#L5-L7

Alexander-Stuckey (changes requested):

talos.nf
MattWellie: I knew there must be something for this exact situation... I found FromFilePairs which creates a list of pairs from a glob... not really what I want

nf_test_inputs/nextflow.config
```diff
@@ -0,0 +1,25 @@
+// Input data paths
+params.input_dir = "/Users/mwelland/code/talos/nf_test_inputs/"
```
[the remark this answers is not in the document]
MattWellie: I meant to have this PR as a draft, accidentally made it not a draft, then forgot how to draft it again 😅

talos.nf
MattWellie: Yee nice, I'd forgotten about this. I have a local modules dir but wanted to get the syntax right before breaking it up

Dockerfile
```diff
@@ -9,16 +9,33 @@ RUN apt update && apt install -y --no-install-recommends \
...
 
 ENV PATH=$PATH:/opt/google-cloud-sdk/bin
 
-COPY requirements*.txt README.md setup.py .
+# Add in the additional requirements that are most likely to change.
+COPY requirements*.txt README.md setup.py ./
+COPY src src/
+RUN pip install --upgrade pip && pip install .[cpg]
+
+FROM base AS talos_none
```
cassimons, line 34: Out of interest, why do you install twice  +/- gcloud?

README.md
```diff
@@ -6,15 +6,18 @@
 
 Talos is a Python variant prioritisation tool which finds variants with diagnostic relevance within large callsets. It incorporates consequence annotation, family structures, participant phenotypes, and existing clinical knowledge to identify variants likely to be relevant to participant phenotypes. It has been configured to do this with high specificity whilst retaining maximal sensitivity to reduce burden on curators.
 
-Analysis consists of two main phases:
+Analysis consists of three main phases:
 
-1. Variant Categorisation
+1. Selection of 'Region of Interest' for the analysis
```
cassimons, line 11: Is "region the right word here? Would Loci be a better fit? 
```suggestion
1. Selection of 'Loci of Interest' for the analysis
```

Dockerfile
cassimons: OR - I could just read the new bits in the README and learn for my self...

cassimons (approved):
Looks good to me!

Dockerfile
MattWellie: Heh. This was something Miah wanted so he didn't have to keep customising the dockerfile to drop the 2GB Gcloud install. There's probably a sexier way to do this as a condition build, but polishing this file felt like a distraction

README.md
MattWellie: Maybe? They're both kinda wrong, as it's really just a gene panel by name based on VEP annotation, I just thought ROI would be the more generic term. I'll swap this out for gene panel to be explicit

talos.nf
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl=2
```
Alexander-Stuckey, line 3: dsl2 should be the default as of Nextflow 22 (I believe), so you don't need this line. Unless you are wanting to support older versions of Nextflow...

talos.nf
```diff
@@ -0,0 +1,104 @@
...
+    publishDir 'results', mode: 'copy'
+
+    input:
+        path matrix_table
+        path panelapp_data
+        path pedigree
+
+    output:
+        path "${params.cohort}_small_variants.vcf.bgz", emit: "vcf"
+        path "${params.cohort}_small_variants.vcf.bgz.tbi", emit: "index"
```
Alexander-Stuckey, line 29: Unless you have a good reason to want to handle these separately, you could emit them as a `tuple` so that the index always remains paired with the VCF

talos.nf
```diff
@@ -0,0 +1,104 @@
...
+include { GeneratePanelData } from './modules/talos/GeneratePanelData/main'
+include { QueryPanelapp } from './modules/talos/QueryPanelapp/main'
+include { FindGeneSymbolMap } from './modules/talos/FindGeneSymbolMap/main'
+include { RunHailFiltering } from './modules/talos/RunHailFiltering/main'
+include { ValidateMOI } from './modules/talos/ValidateMOI/main'
+include { HPOFlagging } from './modules/talos/HPOFlagging/main'
+include { CreateTalosHTML } from './modules/talos/CreateTalosHTML/main'
+
+
+process RunHailFilteringSV {
```
Alexander-Stuckey, line 17: Why is this process still part of the main workflow, rather than being off in a module? Also, is it actually called by anything?

Alexander-Stuckey (commented):
A couple comments

talos.nf
MattWellie: I don't want to explicitly support more than I have to, but this model relies on the user having nextflow installed locally, which could be old or new

talos.nf
MattWellie: Oh yeah, I meant to circle back to this

MattWellie: ah, I didn't update this because I'm not using it anyway. Deleted per the other comment

Alexander-Stuckey (approved):
Looks good to me

PR #460, opened 2024-11-21
Use tenacity as a 3rd party retry wrapper

MattWellie:
# Fixes

  - In FindGeneSymbolMap we sometimes hit errors which crash the process
  - This has been tuned for us (Australia?) but the threshold for query frequency might be different elsewhere... using a backoff wrapper is a better approach than just experimenting with batch sizes until it stops breaking

## Proposed Changes

  - use the tenacity library to retry the queries
  - replace the homebrewed retry/backoff wrapper with tenacity elsewhere as well

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,4 +16,5 @@ python-dateutil>=2
 # this is a ruff package with a python interface, limited to specific python versions
 semsimian
 tabulate>=0.8.9
+tenacity>=9.0.0
 toml==0.10.2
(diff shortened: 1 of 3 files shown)


cassimons (approved):
Looks great. Is `stop_after_attempt(3)` enough retrys?

MattWellie:
I hope so 😬 the wait is 2**attempt seconds, so this allows a final gap of 8 seconds between the last two attempts. No reasons not to boost that a little, or make it configurable...

PR #461, opened 2024-11-22
Fix comp het exclusion

MattWellie:
# Fixes

  - It's been noted that we occasionally let one variant of a comp-het pair slide off the report. This doesn't make sense, as a pairing between two variants should be symmetrical in terms of reaching the report.
  - I originally assumed this was related to the report changes; we no longer create a 'latest-only' report, but when we did there was some logic around "keep all variants which were new in the latest round, and any variants they pair with"
  - The real fault here was that the `check_frequency_passes` method (which permits/blocks variants based on gnomAD frequencies) was usually combined with an escape "always allow if pathogenic in ClinVar".
  - This ClinVar escape only applied to the 'principal' variant in a pair, and not the partner:

```
Variant 1 is common, and pathogenic in ClinVar
Variant 2 is rare, and not in ClinVar
Variant 1 can be the 'principal', as the first frequency test permits ClinVar variants through. 
   - Variant 2 passes as a partner because it's sufficiently rare. 
   - Variant1::Variant2 is a permitted comp-het
When Variant 2 is the 'principal' it's sufficiently rare to pass the test
   - Variant 1 fails the comp-het test as it's too common, and the logic didn't allow for ClinVar status. 
   - Variant2::Variant1 doesn't reach the report.
```

I'm still unsure if this is a changed behaviour, or just a rare occurrence that we haven't paid enough attention to. The logic in this region is unchanged in the last ~7 months, and even then it was restructured, but this bug was from even earlier.

## Proposed Changes

  - Removes the concept of 'singleton' analysis (artificially stripping de novo status etc.). This can be done by providing a singleton pedigree, this is an opportunity to smooth out some logic
  - Changed some more things that SonarLint was shouting at me for
  - `check_frequency_passes` has a new argument - permit_clinvar, defaulting to True. If True, we check if the variant is Pathogenic in ClinVar; if so, the check automatically passes.
  - This removes the need to check "check_frequency_passes or clinvar Pathogenic" elsewhere, bundling both behaviours into the same method

## Checklist

- [x] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/src/talos/CPG/MakePhenopackets.py
+++ b/src/talos/CPG/MakePhenopackets.py
@@ -25,7 +25,7 @@ from metamist.graphql import gql, query
 from talos.static_values import get_logger
 
 HPO_KEY = 'HPO Terms (present)'
-HPO_RE = re.compile(r'HP:[0-9]+')
+HPO_RE = re.compile(r'HP:\d+')
 PARTICIPANT_QUERY = gql(
     """
 query MyQuery($project: String!, $sequencing_type: String!, $technology: String!) {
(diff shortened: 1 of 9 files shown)


cassimons (approved):
Sorry for the delay getting to this.

PR #464, opened 2024-12-10
Remove HPO Recursion

MattWellie:
# Fixes

  - I've encountered some recursion issues in the "find all HPO terms between <this> and the ontology root" search, part of the GeneratePanelData Stage

## Proposed Changes

  - drop the recursion
  - use the `networkx.dfs_successors` method. This is not a class method for the HPO object, but the HPO tree is read as a networkx graph instance, so we can use the method in a slightly different way
  - this method rapidly gets all nodes between the query node and the ontology tree, which is WAY easier to implement than the recursive search
  - this new approach can also search within a finite number of layers - if I'd known it existed I wouldn't have bothered writing it in the over-complicated way to begin with.

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


cassimons (approved):

MattWellie:
This has been tested in a deploy directly from the branch - problems solved in production 👍

PR #468, opened 2025-01-10
Print metadata from report json

MattWellie:
# Fixes

  - Sometimes we'd like to get a quick cohort overview from an old analysis


cassimons (approved):

MattWellie:
@MattWellie to add to this a break down of variant category/frequency

PR #479, opened 2025-02-04
Stricter moi filters

MattWellie:
# Fixes

  - This adds a callset AF filter to dominant variants

## Proposed Changes

  - new method `too_common_in_callset` - this first checks for a minimum callset AC (so we don't over filter in smaller cohorts), then filters out any variants above a given frequency in the callset. this method doens't care about Clinvar status.
  - unrelated change - the `check_frequency_passes` has been inverted to `check_frequency_fails`, then renamed to `too_common_in_population`, which feels much more self-explanatory - now the test when we're excluding variants is `if too_common_in_population` rather than `if not check_frequency_passes` which is just easier to comprehend

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass


cassimons (approved):
Looks good, but it feels like adding a test or two for check_callset_af_fails may not be a bad idea?

PR #485, opened 2025-02-11
Flexible support configuration

MattWellie:
# Fixes

  - We have no way of flexing the importance of any individual category from full to support via configuration

## Proposed Changes

  - removes reference to "categorysupportX" in the documentation, that name prefix is no longer a usable mechanic
  - adds a `support_categories` entry to the config template, this is pulled into each variant as an attribute, and contains the names of all the categories to use with second-tier importance
  - reeeeeeallly simplifies the logical tests. Instead of treating full and support categories separately, and having a test for each combination, this is boiled down to a single method:

1. get all the categories that apply to this variant (Boolean) and this [sample + variant] (Sample) as a set of Strings
2. if support is not being used in this test, subtract all the category names in the `support_categories` list
3. return True if there's anything left in the list

# ALSO

Adds a configurable minimum number of alt-supporting reads for a variant to reach the report

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


cassimons (approved):
Looks great!

PR #486, opened 2025-02-12
Fix partner variant bug

MattWellie:
# Fixes

  - A bug in the support category checking prevented variants reaching the report if... they had other categories assigned, but the partner variant didn't?
  - A weird one for sure, it meant that if a Cat6 only and a Cat1 only were in the VCF, and Cat6 was a support, the only variant which would reach the report was the 6 🙃 

## Proposed Changes

  -
  -
  -

## Checklist

- [ ] Related Issue created
- [ ] Linting checks pass

### DIFF
--- a/src/talos/moi_tests.py
+++ b/src/talos/moi_tests.py
@@ -101,12 +101,12 @@ def check_for_second_hit(
     if sample not in comp_hets:
         return []
 
-    partners = comp_hets[sample].get(first_variant, [])
-
     # thin out the possible partners by alt depth
-    partners = [partner for partner in partners if not partner.check_minimum_alt_depth(sample, min_alt_depth)]
-
-    return partners
+    return [
+        var
+        for var in comp_hets[sample].get(first_variant, [])
+        if not var.check_minimum_alt_depth(sample, min_alt_depth)
+    ]
 
 
 class MOIRunner:
@@ -488,7 +488,7 @@ class RecessiveAutosomalCH(BaseMoi):
                         [
                             principal.sample_category_check(sample_id, allow_support=False),
                             partner_variant.sample_category_check(sample_id, allow_support=False),
-                        ]
+                        ],
                     )
                 ):
                     continue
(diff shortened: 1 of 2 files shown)


src/talos/moi_tests.py
```diff
@@ -372,10 +374,8 @@ def run(
             # force a minimum depth on the proband call
             # and a minimum number of alt reads supporting
             if (
-                not (
-                    self.pedigree.by_id[sample_id].affected == '2'
-                    and principal.sample_category_check(sample_id, allow_support=False)
-                )
+                self.pedigree.by_id[sample_id].affected != '2'
+                or not principal.sample_category_check(sample_id, allow_support=False)
```
MattWellie, line 378: none of this ended up being relevant, but it makes the logic easier to read

src/talos/moi_tests.py
```diff
@@ -101,10 +101,12 @@ def check_for_second_hit(
...
-    partners = comp_hets[sample].get(first_variant, [])
-    if require_non_support:
-        return [partner for partner in partners if partner.sample_category_check(sample, allow_support=False)]
-    return partners
+    # thin out the possible partners by alt depth
+    return [
+        var
+        for var in comp_hets[sample].get(first_variant, [])
+        if not var.check_minimum_alt_depth(sample, min_alt_depth)
+    ]
```
MattWellie, line 109: I think this was the issue - I was getting a Bool value for if the current variant was support-only, then using that to decide if a partner could also be support-only. It was a mess of booleans, so I've dropped it from here completely. Once we find all the partners, there's a simpler test on each pair that at least one must be non-support.

src/talos/moi_tests.py
```diff
@@ -101,10 +101,12 @@ def check_for_second_hit(
...
 
-    partners = comp_hets[sample].get(first_variant, [])
-    if require_non_support:
-        return [partner for partner in partners if partner.sample_category_check(sample, allow_support=False)]
-    return partners
+    # thin out the possible partners by alt depth
+    return [
+        var
+        for var in comp_hets[sample].get(first_variant, [])
+        if not var.check_minimum_alt_depth(sample, min_alt_depth)
```
cassimons, line 108: I know it is not part of this PR, but the `check_minimum_alt_depth` method name is a bit curly (ie the `not` in front of it confused me)

MattWellie: Yup, I think I should revisit all the method names to optimise the logic they're involved in...

cassimons (approved):
These name changes are great.

PR #488, opened 2025-02-17
CPG-Flow migration

MattWellie:
# Migration of the Talos workflow to CPG-Flow land

  - Follow-on from https://github.com/populationgenomics/ClinvArbitration/pull/14
  - Minor syntax improvements flagged by Ruff
  - Adds some files managed by Git-LFS (required for deployment, but until now we've required new users to find and download them separately)

## Proposed Changes

  - Adds `stages.py` containing all the Talos Stage logic, and a `run_workflow.py` implementing a single-pipeline main.py interface (mostly a migration of https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/talos.py)
  - A couple of minor tweaks all over (using `strict=True` when zipping iterables, using dates with timezones, etc.)
  - The whole install has been moved into pyproject.toml, instead or being in setup.py. Saves maintaining both files. 
  - Installation is preferred using UV, with a `uv compile` lock file
     - The Dockerfile and pip-audit CI Workflow have been changed accordingly

# TODO
- needs documentation about the new files & how to mount them in, and a little developer guide if the lock file is going to be kept up to date

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


MattWellie:
This should have been like... 3 PRs, not one

PR #494, opened 2025-03-13
Add custom Gene-MOI data

MattWellie:
# Fixes

  - We don't have a way to overrule PanelApp MOI (e.g. to use Talos in a more exploratory way, testing for new gene~disease associations)
  - We don't have a way to use a non-PanelApp gene data source

## Proposed Changes

  - After all PanelApp data has been queried as normal, checks config to see if any additional gene data should be added (example section in example_config.toml)
  - User must supply at least an ENSG ID with each manually applied gene, optionally a MOI, symbol, chromosome
  - If the gene was already detected via PanelApp, the Custom gene panel is added 
    - extra panel ID 0 associated with the gene, a marker to detect manually updated genes
    - If the PanelApp MOI was different to the manually supplied one, overwrite it. 
  - If the gene was not detected in PanelApp, add a new gene section
  - By default, the MOI for manually supplied genes is mono_and_biallelic, unless set specifically in config

## Checklist

- [ ] Related Issue created
- [X] Linting checks pass

### DIFF
--- a/src/talos/RunHailFilteringSv.py
+++ b/src/talos/RunHailFilteringSv.py
@@ -45,6 +45,14 @@ def rearrange_annotations(mt: hl.MatrixTable, gene_mapping: hl.dict) -> hl.Matri
         same MT, with shifted annotations
     """
 
+    # accept, but don't force, this GATK-SV field
+    if 'ALGORITHMS' not in mt.info:
+        mt = mt.annotate_rows(
+            info=mt.info.annotate(
+                algorithms=['gCNV'],
+            ),
+        )
+
     mt = mt.annotate_rows(
         info=hl.struct(
             AC=mt.info.AC,
--- a/src/talos/ValidateMOI.py
+++ b/src/talos/ValidateMOI.py
@@ -201,6 +201,10 @@ def clean_and_filter(
 
     cohort_panels = set(config_retrieve(['GeneratePanelData', 'forced_panels'], []))
 
+    # add the custom gene panel ID as a forced panel
+    # any gene added manually, either to override MOI, or to add the gene completely, will be applied to every family
+    cohort_panels.add(0)
+
     panel_meta: dict[int, str] = {content.id: content.name for content in panelapp_data.metadata}
 
     gene_details: dict[str, set[int]] = {}
--- a/src/talos/example_config.toml
+++ b/src/talos/example_config.toml
@@ -16,15 +16,30 @@ forced_panels = [1, 2, 3]
 # we treat the gene as new. New/Recent genes are highlighted in the report with a Star next to the symbol
 within_x_months = 6
 
+[GeneratePanelData.manual_overrides]
+# this section permits the manual addition of genes to the panel data
+# each gene here will be folded into the panelapp data after standard API queries have taken place
+# if the gene was already identified for this analysis, the MOI provided here will take precedent
+# if the gene here was not identified through panelapp, a new entity will be added, and the panel ID
+# associated with it will be the number zero (panel IDs must all be numeric, 0 is not in use)
+[[GeneratePanelData.manual_overrides.genes]]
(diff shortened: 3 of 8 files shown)


cassimons (commented):
Any chance we can make this work (easily) so that you can specify new genes, not just new MOIs?  It would be nice to have a config specified candidate gene list.

MattWellie:
> Any chance we can make this work (easily) so that you can specify new genes, not just new MOIs? It would be nice to have a config specified candidate gene list.

Yup, I'll drop something in the config. At minimum you'd need to provide an ENSG and a MOI, optionally a gene symbol (I can source that from the ensembl file), and optionally a list of panels you want it added to, otherwise I'll... lie and say it was on the Mendeliome to it applies to app participants?

cassimons:
yea that all sounds fine. Our of choice I would have it define its own "custom" gene panel that we could surface/flag in the reports somehow

src/talos/RunHailFilteringSv.py
```diff
@@ -45,6 +45,14 @@ def rearrange_annotations(mt: hl.MatrixTable, gene_mapping: hl.dict) -> hl.Matri
         same MT, with shifted annotations
     """
 
+    # accept, but don't force, this GATK-SV field
+    if 'ALGORITHMS' not in mt.info:
+        mt = mt.annotate_rows(
+            info=mt.info.annotate(
+                algorithms=['gCNV'],
+            ),
+        )
```
MattWellie, line 54: unrelated change, but this is currently blocking exome/cnv analysis

src/talos/ValidateMOI.py
```diff
@@ -201,6 +201,10 @@ def clean_and_filter(
 
     cohort_panels = set(config_retrieve(['GeneratePanelData', 'forced_panels'], []))
 
+    # add the custom gene panel ID as a forced panel
+    # any gene added manually, either to override MOI, or to add the gene completely, will be applied to every family
+    cohort_panels.add(0)
```
MattWellie, line 206: by allocating all custom genes to panel 0, we have a way to ensure we never remove these genes from a family's results because the 'custom' panel wasn't allocated to them. Panel 0 is applied to all manual genes, and Panel 0 is applied to every family. We can surface this later in the reports as a manually applied gene

src/talos/example_config.toml
```diff
@@ -16,9 +16,24 @@ forced_panels = [1, 2, 3]
...
-[FindGeneSymbolMap]
-# much higher than this and we get into throttling issues
-chunk_size = 800
+[GeneratePanelData.manual_overrides]
+# this section permits the manual addition of genes to the panel data
+# each gene here will be folded into the panelapp data after standard API queries have taken place
+# if the gene was already identified for this analysis, the MOI provided here will take precedent
+# if the gene here was not identified through panelapp, a new entity will be added, and the panel ID
+# associated with it will be the number zero (panel IDs must all be numeric, 0 is not in use)
+[[GeneratePanelData.manual_overrides.genes]]
```
MattWellie, line 25: This double-bracketed syntax means we can add an unlimited number of these sections to the config file as a list. These can be written manually, or done by writing a list of dictionaries and using `print(toml.dumps(THE_LIST))` on the object.

src/talos/templates/variant_table_row.html.jinja
```diff
@@ -39,7 +39,7 @@
     <td>
         {# Gene symbol #}
         {% for gene in variant.genes %}
-        <a href="https://panelapp-aus.org//panels/entities/{{gene[1]}}" target="_blank">{{gene[1]}}</a>
+        <a href="https://panelapp-aus.org/panels/entities/{{gene[1]}}" target="_blank">{{gene[1]}}</a>
```
MattWellie, line 42: unrelated correction, panelapp links are broken

cassimons (approved):

PR #503, opened 2025-04-01
Alter how filters are configured and applied

MattWellie:
# Fixes

  - Filtering is complicated

## Proposed Changes

  - Adds 4 new filtering classes:
 1. ClinvarDominant - to be applied to Clinvar Dominant...
 2. Dominant - to be applied to all other Dominaint
 3. Clinvar - to be applied to all non-Dominant Clinvar
 4. Global - minimum standard to be applied to everyone

Each of these filters takes specific named parameters from config, so all behaviour can be completely adjusted. 

Each MOI test class uses two of these - a global, and a clinvar filter. All variants are passed though the corresponding global or clinvar-specific tests, depending on whether the gene has a dominant MOI model or not, and whether the variant has a clinvar pathogenic rating.

## Checklist

- [ ] Related Issue created
- [x] Linting checks pass


src/talos/example_config.toml
```diff
@@ -36,25 +36,49 @@ symbol = 'AGENE'
...
 [ValidateMOI]
-# thresholds for different filters during the MOI checks
-gnomad_dominant = 0.001
 
-# by default common in callset/gnomAD filters aren't applied to ClinVar pathogenics
-# set this flag to false to apply apply these filters
-# in testing this created a drastic reduction in candidates, at the expense of a few causative variants
-allow_common_clinvar = true
+# callset AF filtering will only be applied to variants with this AC or higher
+min_callset_ac_to_filter = 10
```
cassimons, line 41: Was this always 10 or did I bump it up, or did it used to be configured differently ant different points in the filters? Not sure what the "right" answer is, just something for us to consider.

MattWellie: I think it may have been 5. I'll pull this into the main page of the report, it would be nice to see at a glance

cassimons (approved):

PR #506, opened 2025-04-08
Flexible hpo terms missing

MattWellie:
# Fixes

  - I forgot to include a run_workflow method for the cpg-flow implementation
  - HPO ~ Panel matching fails if terms are missing in the exact version of the ontology we're using

## Proposed Changes

  - If an HPO term doesn't exist, log a warning and skip over it instead of failing
  - Removes setup.py, no longer required (should have been removed in the previous PR)
  - Adds a cpg-flow workflow runner


cassimons (approved):
LGTM. Do we need to get religious about including a change log from now on?

MattWellie:
> Do we need to get religious about including a change log from now on?

Probably should. I think I'll treat 7.0.0 as the 'start' for changelog purposes, there's just too much stuff prior to that point, and if this is the global starting point for external users there's no need to really tuck into all the details

PR #509, opened 2025-04-15
Update NextFlow for single VCF/normalisation

MattWellie:
# Fixes

  - NextFlow pipeline breaks if the merged file is not normalised
  - BCFtools stage breaks if only one VCF is provided

## Proposed Changes

  - update the pre-merged path to include a `bcftools norm` instead of a `view` to force normalisation (required for echtvar)
  - update the vcf collect method (chatty sold me up the river on that one)
  - `bcftools merge` command includes the `--force-single` command (even though ideally single VCFs follow the pre-merged route


EddieLF (approved):
LGTM although I can't comment on the nextflow .collect() method stuff - this just makes the inputs `ch_vcfs` and `ch_tbis` into list arguments?

MattWellie:
> this just makes the inputs `ch_vcfs` and `ch_tbis` into list arguments?

Yup! This was in place prior to the last PR and it worked, then I had a copilot session to solve something unrelated and it said this wasn't valid. I changed it and it broke 🙃 this is a reversal back to when it worked

PR #513, opened 2025-04-23
Mention AlphaMissense in the `large_files` requirements

lgruen:
Purely a documentation change.


MattWellie:
I've used this open branch as an opportunity to update some other outdated bits of the documentation (relating to compliance with VEP-VCFs, and the way we engage with PanelApp)

MattWellie (approved):
LGTMatt

PR #514, opened 2025-04-23
Fix: Add --no-same-owner flag to tar extraction commands to resolve Docker-to-macOS permission issues

lgruen:
When Docker containers extract tar archives on mounted macOS volumes, permission issues occur because:
1. The container tries to preserve original ownership metadata from the archives
2. macOS refuses these ownership changes when they don't match the host user
3. This happens even with locally created archives due to container-to-host permission model differences

Adding `--no-same-owner` tells tar to extract files as the current user instead of preserving original ownership.

Example error:

```
❯ nextflow -c nextflow/talos.config run nextflow/talos.nf --matrix_tar nextflow/cohort_outputs/cohort.mt.tar
...
ERROR ~ Error executing process > 'RunHailFiltering (1)'

Caused by:
  Process `RunHailFiltering (1)` terminated with an error exit status (2)


Command executed:

  export TALOS_CONFIG=config.toml

  tar -zxf clinvarbitration.tar.gz
  tar -xf cohort.mt.tar

  RunHailFiltering         --input cohort.mt         --panelapp cohort_panelapp_results.json         --pedigree cohort_pedigree.ped         --output cohort_small_variants_labelled.vcf.bgz         --clinvar clinvarbitration_data/clinvar_decisions.ht         --pm5 clinvarbitration_data/clinvar_pm5.ht         --checkpoint checkpoint

Command exit status:
  2

Command output:
  (empty)

Command error:
  tar: clinvarbitration_data/clinvar_pm5.ht/globals/parts/.part-0.crc: Cannot change ownership to uid 502, gid 50: Permission denied
```


MattWellie (approved):
Thanks for this! Not expecting MacOS as a deployment environment, but it's definitely used as a testing one

PR #518, opened 2025-04-29
Re-add PL substitution derived from GQ

MattWellie:
# Fixes

  - Talos is not flexible to missing PL scores in VCFs
  
## Proposed Changes

  - Re-add previous PL annotation, derived from GQ score

General contract:
  - GT/GQ are mandatory FORMAT fields
  - DP is optional, can be replaced with sum(AD)
      - AD is also optional, in which case 'depth' is completely fabricated
  - info.AC is mandatory, as an array
  - info.AF is mandatory, as an array
  - I don't think info.AN is mandatory, but is expected.

Everything else should be provided by the annotation process


cassimons (approved):
Great (but frustrating we need this 🙄 )

PR #524, opened 2025-05-01
Pre download panelapp, adjust Talos to accommodate

MattWellie:
This is #517 again, but on a branch which was rebased in light of separate changes

# Fixes

  - Talos expects to be able to query for PanelApp data at will, which might not be possible in some environments
  - We query for all the information in PanelApp twice per run, instead of once per monthly update
  - PanelApp queries are currently in series, as the prev. server couldn't tolerate high loads

## Results

We now have adequate test data 🎉 so this version has been run on a fragment of the acute-care cohort:

* [code currently on main](https://test-web.populationgenomics.org.au/acute-care/reanalysis/2025-05-01/summary_output.html) (https://batch.hail.populationgenomics.org.au/batches/594411 & https://batch.hail.populationgenomics.org.au/batches/594417)
* [code on this branch](https://test-web.populationgenomics.org.au/acute-care/reanalysis/2025-05-01_predownload/summary_output.html) (https://batch.hail.populationgenomics.org.au/batches/594434)

The current test data has no phenotypic information, which I would definitely like to incorporate into this validation. I hope to be able to do that tomorrow.

## Proposed Changes

  - Adds a new Stage `DownloadPanelApp` - query for all the panel ID~HPO term associations, and all the (green) genes and their MOI on every panel in PanelApp. This data format effectively contains the whole of PanelApp at a given timepoint:
        * every Panel ID, name, disease group, and current version number
        * the HPO terms (if any) corresponding to each Panel 
        * every green gene
        * the panels each gene is green on
        * the MOI on each gene where it is green
        * the date each gene was made green on each panel

  - This PanelApp result file contains a created date, so we can flag to users when they are using reference data which should be updated.

  - Adds a new Stage UnifiedPanelAppParser. This does the combined work of the previous 2 stages `GeneratePanelData` (use participant phenotype terms to get the panels of interest per participant) and `QueryPanelApp` (using the list of panels, query for their content). This stage is offline, using the cached data from `DownloadPanelApp`. This Stage exports a single file containing both the participant~panel associations, and the genes in this analysis with all relevant information.

The remaining ~1000 lines of this PR is a mass implementation of this new result:

- NextFlow implementation uses the new functions
- cpg-flow uses new stages (generating new download/cache once per month, then re-using it for every run & cohort)
- instead of passing both the PanelApp and participant-panels files separately downstream, a single file contains both blocks of data, for a cleaner interface
- two previous panelapp modules and corresponding tests deleted - replaced with new modules and tests
- extensive re-working of the panel/phenotype elements in ValidateMoi in response to new data structure

### A couple of incidental behavioral changes 

- there was some logic to squash all reportable events at a single position into one row of the output HTML (both `same position + gene, multiple MOIs`, and `same position but separate genes`). I'm not sure if the `same position but separate genes` squashing was ever triggered outside of tests, but it's a terrible idea. There are a ton of attributes which should be kept in parallel (transcript consequences, gene panels, etc.), so I've removed all this squashing to make things easier. This will mean some light duplication of variant in the HTML, but each row will be a distinct reportable event (AD vs. recessive etc.). This also makes the labelling easier, e.g. an ambiguous flag for dominant MOI in a dominant/recessive gene is now properly applied to a dominant event, instead of applying to a dominant event, and popping it back off again if the same variant could also be recessive. 

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


src/talos/DownloadPanelApp.py
```diff
@@ -0,0 +1,417 @@
...
+- once to get the gene data for each panel
+
+There's a couple of ways in which this is wasteful:
+- PanelApp is only updated once per month, but this querying goes on in every run
+- the original API query implementation was in series, not parallel, due to server instability, so it was slow
+
+In addition to the above, the core assumption here is that deployment environments will have an internet connection.
+That is typically not true of HPC environments, making this a tough workflow to run in those environments.
+
+This script queries for the whole of PanelApp in one go, using a barrage of highly parallelised queries:
```
cassimons, line 14: We need to discuss this with Simon. Fine while it is a couple of users, but he might want to provide an endpoint for a nightly pre-rendered file for direct download or similar to avoid the load.

cassimons (approved):
😮‍💨 
LGTM, but we should chat about the DL with Simon.

src/talos/DownloadPanelApp.py
MattWellie: Yup, that makes sense. This implementation is the same overall amount of querying as a single run, but it's way more highly parallelised so the point burden could be much harder. So far so good, but I'm the only user 😅

PR #529, opened 2025-05-15
Restore .zip ending for params.gnomad_zip

lgruen-vcgs:
Tiny config fix.


MattWellie:
We're content keeping this as-is for the minute. The annotations file is a `.zip` by default, but we found that users would unzip the file (which effectively breaks the installation). We've removed the extension in the zenodo record and config to prevent people trying this. It's a little dumb, but it's worked so far 😅

PR #530, opened 2025-05-15
Update config as cohort input MTs aren't tar'd up anymore

lgruen-vcgs:
No functional change, just updating README + wiring.


MattWellie (approved):
@MattWellie to follow with alterations to move mode

PR #532, opened 2025-05-29
RD-122: Persistent urls

MattWellie:
# Fixes

  - Closes https://cpg-populationanalysis.atlassian.net/browse/RD-122

## Proposed Changes

  - Adds an extra stage in parallel with the existing Report uploading Stage. This writes the same output to a generic path, instead of a date-specific path, so that we can maintain updated reports at a static URL

## Unrelated

  - move to module level imports (more inline with the google style guide, tidier import blocks)
  - move the bump2version config into pyproject.toml (identical behaviour)


cassimons (approved):
Legend thank you!

PR #537, opened 2025-06-16
Move targets.Dataset -> dataset corrections

MattWellie:
# Fixes

  - improper string replacement

## Proposed Changes

  - reverts dud change, removes unnecessary quotes round type annotations


Alexander-Stuckey (approved):
Looks good to me

PR #560, opened 2025-07-22
fancify downloader

folded:
- **merge and insert WT instead of missing**
- **big README changes**
- **linting**
- **some fixes and doc updates**
- **tons of doc changes**
- **tweaks, removing empty files**
- **Update Configuration.md**
- **Update Configuration.md (#558)**
- **update zenodo record**
- **lots and lots and lots of import block reformatting**
- **sync up test config**
- **corrections to manually added gene content**
- **MORE DETAILS**
- **lint**
- **stay away from Hail 135**
- **allow for an easy ignore**
- **oneshot download script**
- **add in a ref genome**
- **add in a ref genome**
- **mild config changes**
- **Make large file downloader script more fancy.**


folded:
LMK what you think & whether it works for you.

The change to the HG38 URL should happen regardless, but I was lazy and didn't split it into two PRs.

MattWellie:
I've tried running this download (great) and using it (not so great) - the HG38 reference genome from the Broad public bucket isn't actually validly formatted, which was a fun discovery. Samtools some strict formatting rules which means that FASTA can't be indexed, because it contains erratic line length changes (see [here](https://github.com/samtools/samtools/issues/929)). This issue/fix way predates the data in the bucket, so I don't understand what it's doing there.

PR #561, opened 2025-07-22
Allow unknown sexes in pedigree, allow de novo skipping

MattWellie:
# Fixes

  - BCH issue [here](https://centrepopgen.slack.com/archives/C08H4B6PPL4/p1753220374820279) - 'unknown' is a valid enumeration in the pedigree parser, but not tolerated in a dictionary lookup when tallying sexes of participants. This is one to be aware of for the pedigree validation plans.
  - BCH issue [here](https://centrepopgen.slack.com/archives/C08H4B6PPL4/p1752785302497149) - incorrect number of AD fields. This is a mitigation (skipping the de novo calling completely), as the underlying issue looks to be a data problem and will be hard to diagnose remotely

## Proposed Changes

  - allow all unknown sexes to be counted as 'unknown', by changing the strict dictionary lookup into a `.get` with a default of unknown
  - allow de novo calling to be skipped via config, which is a short term mitigation for the observed data format issues. We previously had the ability to ignore this category, but we don't have the ability to not run the evaluation, and that's where data formatting issues are being encountered.

## Unrelated

 - Inside the de novo calling method we used to filter the variants of interest down to `high transcript consequences and spliceAi`. Now that we don't have spliceAi, this is damaging transcript consequences only. I swear this used to also include ClinVar pathogenic, but that's not presently in the code. This tweak adds ClinVar P/LP (any * Stars) to the variants we're interested in detecting de novo inheritance patterns for
 - pin Hail to 0.2.134 (135 introduces terrible QOB behaviour, which we are not prepared for)

## Checklist

- [x] Related Issue created
- [x] Linting checks pass


src/talos/RunHailFiltering.py
```diff
@@ -517,8 +507,11 @@ def filter_by_consequence(mt: hl.MatrixTable) -> hl.MatrixTable:
         ),
     )
 
-    # filter out rows with no tx consequences left, and no splice cat. assignment
-    return filtered_mt.filter_rows(hl.len(filtered_mt.transcript_consequences) == 0, keep=False)
+    # filter out rows with no high impact tx consequences, and no clinvar Pathogenic annotation
+    return filtered_mt.filter_rows(
+        (hl.len(filtered_mt.transcript_consequences) == 0) & (mt.clinvar_talos == 0),
```
cassimons, line 512: I agree we should retain clinvar variants, but why did we stop retaining sAI variants? We really want to be able to use this cat as an option.

cassimons (approved):
Looks good, but I am keen to discuss (internal) support of sAI again.

src/talos/RunHailFiltering.py
MattWellie: We stopped including SpliceAi because we were previously getting all those annotations from the `seqr_loading_pipelines` annotation resource, which wasn't realistic to distribute. The bones of the category still exist in the codebase, so if we have a more portable source for that annotation we can add it back in. That'll mean anyone able to use it without violating license terms can include it, but it'll be optional for everyone else

PR #584, opened 2025-08-21
Automatically download seqr mapping

MattWellie:
# Fixes

  - I have to go and download the CPG ID ~ Seqr ID mapping each time that changes. That process took so long that I'd sometimes go months without updating all the links, meaning Talos reports just didn't link to seqr for new samples

## Proposed Changes

  - Talos has a 'make a runtime config' step, which pulls components from the global config (including parts from the talos config file in the prod-pipes-config repo) into a new file, which it saves into the run's output folder
    -  That prod-pipes-config file has String templates for each project*sequencing type, which are used in Talos to build links to seqr
    - It also contains a path to a file which contains the CPG ID ~ Seqr ID mapping, which I've generated by hand
 
Using the endpoint @EddieLF put into Metamist (awesome):

 - The 'make a config file' step now goes goes directly to metamist. If the `dataset-sequencing_type-long/short-read` combination has a project in seqr, it gets the name of the project and uses it to build the URL templates
 - It goes to the another endpoint and gets the CPG~Seqr ID mapping as a CSV, saves that into the output folder for the run, and adds the path into the config
 - While I'm at it, this process generates path for the 'historic run data' folder too. That used to be a fixed path in the dataset's `test-analysis` bucket (I don't know why I chose to do that). Now it's going to be namespaced in the main dataset bucket on a path like `dataset-bucket/talos/[exome/][long-read/]history`. This will require the latest history file being moved to that new location for each project... I'm not pleased about that, but this is a much more defensible forward-thinking location for it to be. 

## TODO

The mapping from seqr is Family ID : Seqr Family ID, and I need it indexed on CPG ID. Should be easy enough to update this.

## Checklist

- [ ] ChangeLog Updated
- [ ] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass

### DIFF
--- a/src/talos/CreateTalosHTML.py
+++ b/src/talos/CreateTalosHTML.py
@@ -496,7 +496,7 @@ class LinkEngine:
             external (bool): if True, embed/lookup external IDs in the lookup dictionary. Default is sample.name.
                              This is mostly for a CPG internal use-case, where the seqr lookup and external lookup come
                              from different sources. The Lookup variable makes this redundant.
-            lookup (dict): optional, a path to a JSON dictionary. Can be used to connect sample ID -> arbitrary ID
+            lookup (dict): optional, a path to a CSV/TSV/JSON file, used to connect sample ID -> arbitrary ID
         """
         self.template = template
         self.variant_template = variant_template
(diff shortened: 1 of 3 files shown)


EddieLF (approved):
👍 💯

PR #588, opened 2025-08-26
History is a DB

### LINKED ISSUE #557: Reanalysis / Prior-run input
Hi all, 
The readme mentions that talos consults the history of previous runs to incrementally update the `first_seen` and `evidence_last_updated` dates. I have not been able to figure out exactly how this works: do I need to specify the path to the previous analysis as a parameter? Or is talos expecting it to already exist in a certain location? 
Thanks!

### LINKED ISSUE #587: History file(s) should be a single database
Instead of a collection of JSON files which are read, updated, written, repeat, the full content of a run's history should be contained in a single database.

MattWellie:
# Fixes

  - The current process being used by Talos to communicate 'state' (variants seen in previous runs) is a collection of JSON files. This process works for us (CPG), but doesn't work at all with NextFlow users
  - The issue is with how Talos is expecting to read a path/folder from config, and read their contents in real time, which works perfectly fine on local runs or in the cloud, but will not work in containers. Any files not explicitly read/mounted into the container are not expected to be usable, as the dockerised runtime won't have the ability to reach those paths. 
  - Instead we need to remove any file paths from Talos' config.toml, instead migrating those paths to the NextFlow config file, where they will be read into the working directory.
  - The two remaining parts to changes here are the history directory (Talos expects to list the contents of an arbitrary directory, find the latest file, read it, and then write a new file back) and the seqr-lookup file (expects to read the contents of a file which wasn't copied into the working directory, and use its contents)

Closes #557 
Closes #587 

## Proposed Changes

  - This completely removes the previous record of 'state' using JSON files, moving to a single database (and its copies)
  - This database is very simple, SQLite, with four tables, see schema discussion below.
  - The file containing a mapping for samples to Seqr IDs is now passed explicitly into NextFlow WFs as a file, instead of expecting it to be reachable at runtime from an arbitrary location.
  - NF modules are updated to pass in the database file (optional) and seqr_ID_mapping (optional)

## TODO

  - still need to integrate this into Talos (passing in DB files/copying back out)
  - after merging I'll need to convert the previous history files to SQLite DBs

## db Schema

talos_state_er

For the history, we want to record:

 - each variant we've previously seen in this cohort
   - fairly self explanatory. I toyed with the idea of having the `chr-pos-ref-alt` String _as_ a key instead of needing a whole table, but it feels like this could have other uses, e.g. a really clean way of counting how many unique variants have been seen across the whole cohort.
 - each sample-variant-category combination, and the first time it was seen
   - this is essential for finding which categories have previously been assigned, and getting the earliest date an assignment was made. Using previous dates to make sure old variants are given a lower priority is the core of what Talos's _reanalysis_ process is
 - for each sample-variant combination, the number of associated ClinVar stars, and the date (if appropriate) that the variant had a phenotypic match to the participant
   - both of these fields are dependent on sample-variant, but are not connected to each other or the category. I put these here, but this could easily be two separate, trivially small tables.
   - in the report we annotate each variant with `clinvar_increase` (it was previously pathogenic, and now is pathogenic with more evidence), and `date_of_phenotype_match` (can be None, earliest date this variant was prioritised for the participant). This table tracks both these things, though the two attributes would be updated at separate times.
   - I'm talking myself into making this two separate tables.
 - for each sample-variant combination, a list of any comp-het partners it was seen with
   - this is used to identify when a variant category was seen before, but a new comp-het partner represents new information, so we would highlight this on the report by changing the `evidence_last_updated` attribute to _today_.

Other than the `var_stars` table, I'm confident this is in 3NF - all non-key fields depend only on the key, and there's no repetition within the tables. I even think the `var_stars` table is technically well normalised, it's just dirty.

## Checklist

- [x] ChangeLog Updated
- [ ] Version Bumped
- [x] Related Issue created
- [ ] Linting checks pass


src/talos/convert_json_history_to_db.py
```diff
@@ -0,0 +1,98 @@
...
+        except ValueError:
+            logger.info(f'File {filename} did not have a valid date, skipping')
+
+    if not date_files:
+        logger.warning(f'The folder {results_folder} was provided, but did not contain any valid files')
+        return None
+
+    return str(date_files[max(date_files.keys())].absolute())
+
+"""
```
EddieLF, line 98: What's the purpose of this?

EddieLF (approved):

src/talos/convert_json_history_to_db.py
MattWellie: Ah yeah this shouldn't have been committed - this is the logic I was using the find the most recent history database file. I'll need to use the same to find the latest history JSON, to convert that into a SQLite db.

MattWellie:
this was a fun little hack, but I'm coming around to going further... instead of maintaining a history and results separately, record results into a db and remove the need to separately track two types of file.

PR #590, opened 2025-09-02
Everything is a db... 

MattWellie:
This isn't a full implementation, it's just kicking tires on SQLalchemy, and sketching out a database design for discussion

# Purpose

  - To date Talos has maintained two main objects:
    - a set of JSON results, representing the outcome of a single Talos run
    - a set of History files, representing the outcome of previous Talos runs

This is messy, causes very specific problems around containerised infrastructure, and is hard to move on from (e.g. if we want a web app to present the data in a scalable reactive way we need to move on from JSON files, and if we want to enable feedback on individual variants, that would require an additional object/db to be created)  

#588 was a PR addressing the first part of that problem: turning the history files into a database, but it doesn't go far enough. That still leaves us with history, results, and potentially feedback in 3 separate data stores.

This suggests going further, implementing the history, results, and feedback mechanism into a single unified database. The Schema makes space for all the things we currently store in JSON (participants, family structures, variants, variant annotations, categories and their dates of first assignment, ...), and makes room for some more (a table to record dates/software versions for when Talos was run, a table for logging decisions made by analysts, allowing a participant to be 'solved' by linking a reported event with a proband and a free text field, or a variant to be marked as irrelevant/artifact).

The dream here would be a single file (if a freestanding sqlite db) or a database (if we up our game), where Talos would incrementally update with new findings, without having to write a whole new set of results with each run. No need to track a history separately, the dates are all in the main database. No need to write a second report after the HPO matching is done, we just update the variants we already found with phenotype matches. The Jinja templates can be populated directly from the database, or at some point we could shift to javascript in the web page interacting directly with the database.

Instead of redundantly storing the same variant annotations with each instance of the variant, this has the potential to shrink the overall data footprint of the JSON reports, whilst being far more searchable. 

This version is implemented in SQLite, which saves on the complexity of running a DBMS either in the container or externally. That could be altered at some point in the future. 

Instead of naked queries it's implemented in SQLalchemy (a change to a proper DBMS wouldn't require a change in (much) code). That also enables the embedding of python objects (dicts, sets, lists) as attributes using the PickleType.

SQLite doesn't have a notion of migrations, so we'd have to keep track of any changes to this model. Worst case scenario involves creating completely new tables and schemas, reading old data -> new table, and swapping the table to the original name. A better database tech would have the ability to update/alter tables in place, but this doesn't currently need those features.

  - Schema for consideration and development, does this include everything a Talos.db needs for now? 

## Proposed Changes

  - an SQLite/SQLAlchemy database schema
  - create_sqlite_schema.py contains the schema, defined in terms of SQLalchemy classes and their relationships
  - create_sqlite_db.py is an example of the database being instantiated if it doesn't exist, then some insertions/queries/updates on the contents using sqlalchemy object methods

The create_sqlite_db.py can be run locally with just sqlalchemy as a dependency, feel free to kick it about and see how it feels

my_data_model_diagram

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -45,14 +45,14 @@ This document explains the configuration options, the default values, and sensib
   * Config Section: `RunHailFiltering.de_novo`
   * Description: Controls filtering thresholds and de novo variant detection parameters during the Hail Stage
 
-| Field            | Purpose                                                                                                                                                         | Default | Alternative |
-|------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|-------------|
-| `min_child_ab`   | Minimum proportion of alt-supporting reads relative to overall read depth for the proband                                                                       | 0.20    |             |
-| `min_depth`      | Minimum read depth for calls. In order of preference/availability this is from `DP`, total of `AD`, or if both are absent, a dummy value which will always pass | 5       |             |
-| `max_depth`      | Maximum read depth for calls. In order of preference/availability this is from `DP`, total of `AD`, or if both are absent, a dummy value which will always pass | 1000    |             |
-| `min_alt_depth`  | If all other conditions succeed, at least this many Alt observations to confirm a de Novo call                                                                  | 5       |             |
-| `min_proband_gq` | Minumum genotype quality score for all variants in probands/affected participants (detected via Pedigree)                                                       | 25      | 40          |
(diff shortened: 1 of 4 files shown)


MattWellie:
Graph generated with https://github.com/Dicklesworthstone/sqlalchemy_data_model_visualizer

MattWellie:
distraction for now

PR #594, opened 2025-09-04
Dragen pad homref

MattWellie:
# Fixes

  - Observed with a collaborator - some callers (specifically DRAGEN, possibly limited to DRAGEN + an unusual set of parameters) have a condensed representation for FORMAT.AD. Instead of being a 2-length list for `[REF, ALT]`, the values are instead `[REF]` for HomRef sites. This conflicts with the values we expect in the de novo filtering implementation

## Proposed Changes

  - Adds a new config entry `pad_homref_ad` (this isn't necessarily a data structure we expect to see often, so it shouldn't be a default behaviour)
  - if `pad_homref_ad==True`, we scan the MatrixTable and detect HomRef sites with `hl.len(mt.AD)==1`, and append a zero (representing zero Alt reads) to the AD array. 
  - For any other sites (not HomRef + len(AD) == 1) we leave this unchanged

## Todo

Still requires a test case

## Checklist

- [ ] ChangeLog Updated
- [ ] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass


cassimons (approved):
Looks good.

Fine with it being off by default, but is there any way to surface a helpful error message if we see data that looks like this?

PR #605, opened 2025-09-21
Restore single file HTML report with child-rows

cassimons:
# Fixes

- Restores single file cohort HTML report with in-page sample detail views as child rows.

# Proposed Changes
 - Remove per-sample HTML generation 
 -  Pre-serialize nested variant/sample data for safe Jinja embedding.
 - Replace individual/sample report templates with a single DataTables-powered index that renders expandable child rows client-side.

## Checklist

- [ ] ChangeLog Updated
- [ ] Version Bumped
- [ ] Related Issue created
- [ ] Tests covering new change
- [ ] Linting checks pass

PR #607, opened 2025-09-22
Refactor-report

cassimons:
PR stacked on #605

# Fixes

  - Refactors the report to improve usability

## Proposed Changes

  - Removes many columns to reduce width
  - Rearranges the layout to make better use of space
  - Reintroduces summary and metadata tabs, and adds a run configuration tab
  - Adds an explicit filter for phenotype match
  - Adds a help section
  - Numerous small tweaks

## Checklist

- [x] ChangeLog Updated
- [x] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass


src/talos/CreateTalosHTML.py
```diff
@@ -247,10 +248,20 @@ def get_summary_stats(self) -> tuple[pd.DataFrame, list[str], list[dict]]:
         which passed through the MOI process, not the absolute number
         of variants in the report
         """
-        ordered_categories = ['any', *list(config_retrieve('categories', {}).keys())]
+        categories_config = config_retrieve('categories', {}) or {}
```
MattWellie, line 251: I'm keen to move away from this being in config. Category discovery by the categories in a given dataset looks robust enough, and prevents the need to keep this config entry updated.

I think that means the lines following this can be handled using a defaultdict implementation

MattWellie: Ah, memory coming back. I re-implemented this whole summary in to the JSON [here](https://github.com/populationgenomics/talos/blob/main/src/talos/utils.py#L820-L877) while the stats tab was missing, so this is pre-populated in the JSON

MattWellie: There is some redundancy here, but who cares

PR #618, opened 2025-11-27
Fix de novo calling for dragen

MattWellie:
# Fixes

  - Our collaborators using DRAGEN VCFs & non-Hail combining have not been able to detect de novo variants in their datasets.
  - Upon investigation it appears that the HomRef data we generate internally, which Talos has been designed around, has a format where homref calls all have a depth (DP) and AD attribute, AD is missing but depth is present, e.g.:
  ```
GT:GQ:DP:PID:PG:PS:SB:RGQ:MIN_DP:GP:PGT:AD:PL
0/0:38:33:.:.:.:.:.:.:.:.:.:.
0/0:32:33:.:.:.:.:.:.:.:.:.:.
0/0:44:28:.:.:.:.:.:.:.:.:.:.
```

  - Collaborators providing example data generated from DRAGEN & a different joint calling process have data where HomRef genotypes have an AD annotation but the value is `missing`, and there is no Depth/DP annotation:
  
```
GT:AD:GQ:RGQ
0/0:.:40:.
0/0:.:40:.
0/1:16,13:38:74
```

  - the de novo process as written uses Depth as a proxy for minimum evidence, and went through this algorithm:
    - if DP is present in the schema, use it
    - if DP is absent but AD is in the schema, use the sum of the AD value as DP
    - if neither AD or DP are present, insert a dummy value as DP
    - then do some filtering on minimum depth
  - this broke, because AD was present in the schema, but had a missing value, so the value assigned to 'DP' for HomRef calls would be either 0, or more likely `hl.missing(hl.tint)`, either of which would fail the depth test, so HomRefs would be removed, leaving no parental calls if they were WT.
  - A side note here is that joint-calls generated by DRAGEN's IGG combiner has a _fun_ little quirk, whereby the AD field which should always contain depth counts for [ref, alt] respectively will typically just have a single entry for HomRef calls. This was also able to trip up the de novo checks

## Proposed Changes

  - Creates a new test case `tricky_de_novo` in the test data. This is a de novo inheritance pattern, with parent genotypes missing and no AD populated. There were a couple of changes to the generation script to enable this. This data appears to reflect the data format seen at our collaborators' sites.
  - The test data has been regenerated with two new tricky de novo variants added
  - A new data normalisation step is added to the de novo method: 
    - GT is updated - where a missing genotype is high quality, insert a HomRef call. Unsure about this one. It shouldn't typically be relevant, but in datasets formed through gVCF merging instead of joint-calling the ability to call de novos specifically will be dependent on the caller used.
    - AD is updated/inserted for HomRef variants:
      - If a variant is HomRef and high quality:
        - If AD is populated, check if it has only one AD value (seen in AGG data) - append a second, 0, representing alt counts
        - if AD is missing, add a dummy value of `[min_depth + 1, 0]` - this will prevent depth tests from failing
    - Create a DP (depth) annotation for each genotype. 
      - if DP exists in the schema, use it, otherwise generate a DP annotation from the ref and alt depths

 - The end product here is a process which checks the GT, DP, and AD fields, then updates or inserts values according to what can be inferred from other values. This happens prior to the de novo test running, so sparse genotyping (most samples missing instead of 0/0) or data-saving VCF formats (most annotations trimmed or absent for HomRef sites) no longer causes the de novo detection to fail

This is evidenced by the two test case variants, both de novo, and both having mostly missing data in parents. When run through the dummy workflow one is now detected as a De Novo, and the other is not in a green gene. 

## Semi related

  - `AnnotateCsqWithBcftools` Nextflow module has the `--unify-chr-names` flag set, as addressed in https://github.com/populationgenomics/talos/pull/615
  - updates the download link to the latest version of ClinvArbitration

## Checklist

- [x] ChangeLog Updated
- [x] Version Bumped
- [ ] Related Issue created
- [x] Linting checks pass

### DIFF
--- a/src/talos/RunHailFiltering.py
+++ b/src/talos/RunHailFiltering.py
@@ -599,7 +599,7 @@ def annotate_category_de_novo(
     min_all_sample_gq: int = de_novo_config.get('min_all_sample_gq', 19)
 
     # we've seen some pretty weird data formats, so a 'genotype-only' de novo detection option is being added
-    # the noise reduction offered by using the more complex filtering strategies have shown great performance internally,
+    # the noise reduction offered by using the more complex filtering strategies have shown great performance internally
     # but if the data you have doesn't allow for those to be used, consider using this flag
     genotype_only: bool = de_novo_config.get('genotype_only', False)
 


cassimons (approved):
LGTM. 🤞 this puts this to bed...

PR #626, opened 2026-01-12
Always download the latest ClinvArbitration data

MattWellie:
# Fixes

  - Currently the codebase is set up to download the ClinvArbitration data once upon initial set up using the [large files download script](https://github.com/populationgenomics/talos/blob/main/large_files/gather_file.sh#L140-L142)
  - [A startup check](https://github.com/populationgenomics/talos/blob/main/src/talos/StartupChecks.py#L287-L290) will fail if the downloaded data is > 2 months old - this check makes sense, but is a PITA for users as the process to download a new one isn't clear.
  - I've partially solved this issue before in TalosAf [here](https://github.com/populationgenomics/talos_af/blob/f05544c7572778d0cc27a5a70acc3457208ffe46/src/talos_af/utils.py#L498-L522), but haven't yet tried to implement that in NextFlow

## Proposed Changes

  - Instead of storing a hard link to an exact ClinvArbitration file, we give the numerical ID of the ClinvArbitration record (e.g. for `https://zenodo.org/records/18218150` - 18218150)
  - First new pipeline step - using the Zenodo API, query for the latest version of that record, return the ID
  - Dynamically put the returned ID into a file Path, and check if it exists (e.g. `<large_files>/clinvarbitration_18218150.tar.gz`)
      - If the file exists (previously downloaded), use that file as the workflow's clinvar input
      - If the file doesn't exist, use the second new pipeline step - take the Zenodo ID, use it to get a URL for the latest record's download link, and pipe that into `wget` to download the file

I ran into issues pretty early on trying to apply conditional syntax predicated on a file existence check. My first naive shot at this looked like:

```
GetLatestClinvArbitrationId(params.clinvar)

if file("${params.large_files}/clinvarbitration_${GetLatestClinvArbitrationId.out}.tar.gz").exists() { 
    ch_clinvar = channel.fromPath("${params.large_files}/clinvarbitration_${GetLatestClinvArbitrationId.out}.tar.gz")
} else {
    ch_clinvar = GetLatestClinvArbitrationFile(GetLatestClinvArbitrationId.out)
}
```

It turns out that `if ... {` and `file("{variables}").exists()` logic is applied at compile time, so it didn't work at all with generating on-the-fly file paths using the value returned from `GetLatestClinvArbitrationId`, so I recruited Gemini.

This now makes use of some NF syntax which blew my little mind. Specifically:

```
// 2. Route the ID based on whether the local file exists
ch_id_route = GetLatestClinvArbitrationId.out.map{ it.trim() }
    .branch { id ->
        def targetFile = file("${params.large_files}/clinvarbitration_${id}.tar.gz")
        exists: targetFile.exists()
            return targetFile
        download: !targetFile.exists()
            return id
    }

// 3. Only run the download process for the 'download' branch
GetLatestClinvArbitrationFile(ch_id_route.download)

// 4. Combine them back: use the existing file OR the newly downloaded one
ch_clinvar_tar = ch_id_route.exists.mix(GetLatestClinvArbitrationFile.out)
```

This is mapping the output of `GetLatestClinvArbitrationId` (plus a newline.trim() on the String) into a `.branch` closure. This applies asynchronous logic to the resulting file path, and has two labelled output branches.

The next step only executes at all if the `!exists` branch was selected. The final output channel is then generated by `mix`ing the previously-existed (file) and didn't-exist (fresh download) channels. 

```bash
executor >  local (2)
[8b/276a0f] process > GetLatestClinvArbitrationId   [100%] 1 of 1, cached: 1 ✔
[f0/2e4e1d] process > GetLatestClinvArbitrationFile [100%] 1 of 1 ✔
[ab/158c62] process > StartupChecks (1)             [100%] 1 of 1 ✔

# second run
executor >  local (7)
[32/97aaca] process > GetLatestClinvArbitrationId   [100%] 1 of 1 ✔
[-        ] process > GetLatestClinvArbitrationFile -
[14/4fc15c] process > StartupChecks (1)             [100%] 1 of 1 ✔
```

## As-yet unresolved

- First, we'll need to get people to use this change before their 2-month grace period is up. There is a config option to disable this clinvar date check, but we really want people to be using the latest & greatest.
- Second, we'll either want to tie the generation of the latest monthly data, and then its subsequent upload to zenodo, to a routinely triggered job, or we want to remove the zenodo connection as a requirement.

The second is spicy, I'm actually not sure if a new version of an existing zenodo record can be created entirely via the API. If we can't do that, we're stuck having to pump out a new version of the ClinvArbitration data each month or two, otherwise even this automated download will fall foul of the download-latest rule.

Currently we're doing this on a routine basis for our internal use of the tool, but it's something worth considering in terms of long-term support. This is still a more user friendly version of what we have, but it's more like a really good band aid than a solution.

A _solution_ would be installing/cloning the ClinvArbitration source code into this repository and calling it on a regular basis to make the first step of the Talos workflow:

 - do we have a file called `<Year>_ <Month>_Clinvarbitration.tar.gz` matching the current year and month?
   - if so use it
   - if not, make it

That would probably be easier than what I've done here, and more sustainable, but leads to the duplication of effort (every site downloads the whole of Clinvar every month).

Basically I was thrilled at the nextflow syntax here, and even if we don't go with this solution, I wanted y'all to look at it. It's cool.

## Checklist

- [ ] ChangeLog Updated
- [ ] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,5 +4,7 @@ build
 .github
 .coverage*
 .nextflow
+.git
 
 large_files
+nextflow
--- a/large_files/gather_file.sh
+++ b/large_files/gather_file.sh
@@ -137,10 +137,6 @@ start_download https://github.com/obophenotype/human-phenotype-ontology/releases
 
 start_download https://github.com/obophenotype/human-phenotype-ontology/releases/download/v2025-05-06/genes_to_phenotype.txt
 
-# latest clinvarbitration data
-CLINVAR="clinvarbitration.tar.gz"
-start_download https://zenodo.org/records/17577130/files/ClinvArbitration_Nov_2025_clinvar_decisions.release.tar.gz?download=1 "${CLINVAR}"
-
 # AlphaMissense raw data
 AM="AlphaMissense_hg38.tsv.gz"
 start_download "https://zenodo.org/records/8208688/files/AlphaMissense_hg38.tsv.gz?download=1" "${AM}"
--- a/nextflow/modules/talos/GetLatestClinvArbitrationFile/main.nf
+++ b/nextflow/modules/talos/GetLatestClinvArbitrationFile/main.nf
@@ -1,0 +1,17 @@
+process GetLatestClinvArbitrationFile {
+    publishDir params.large_files, mode: 'copy'
+
+    container params.container
+
+    input:
+        val current_zenodo
+
+    output:
+        path "clinvarbitration_${current_zenodo}.tar.gz"
+
+    shell:
+    '''
+    download_link=$(python -m talos.check_clinvarbitration_zenodo !{current_zenodo} --download)
+    wget "${download_link}" -O clinvarbitration_!{current_zenodo}.tar.gz`
+    '''
+}
--- a/nextflow/modules/talos/GetLatestClinvArbitrationId/main.nf
+++ b/nextflow/modules/talos/GetLatestClinvArbitrationId/main.nf
@@ -1,0 +1,13 @@
+process GetLatestClinvArbitrationId {
+    container params.container
+
+    input:
+        val current_zenodo
+
+    output:
+        stdout
+
+    """
+    python -m talos.check_clinvarbitration_zenodo ${current_zenodo}
+    """
+}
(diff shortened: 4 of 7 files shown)


EddieLF (approved):
LGTM

PR #632, opened 2026-01-21
Just a de novo patch

MattWellie:
# TL/DR

During the previous de novo fix I solved issues for some collaborators, but it's still broken for others. Based on local tests the code requires either DP or AD to be populated with real values. Where DP is absent from the schema, and AD contains missing values, all HomRef entries are being thrown away.

### Explanation

Talos' de novo detection previously used `GT`, `AD`, and `DP`. Allowing for sparse/combined-gvcf callsets:
* If AD was missing but the genotype had a high `GQ`, I chucked in a fake `AD` value (`[min_depth + 1, 0]`)
* Then if DP was missing I created it from `sum(AD)`, which could have been the artificial values

This caused mass failures because some logic was looking at the second value of the AD array to get an alt-depth count, but some DRAGEN-called data only has one `AD` value in the array. Ironically on the tiny test VCF sent through to troubleshoot, this all worked first time on main, so I primarily used data from another collaborator to diagnose the issues.

During the change I removed most references to `AD`, instead deferring those checks until later. I left in some logic which generated DP from AD, which meant:
* I was no longer populating `AD` with dummy values
* I was still generating `DP` from `sum(AD)`

In the data sent over, `DP` didn't exist in, and `AD` was only populated for non-ref calls. This threads the needle so that HomRef samples are all given a missing value for `DP` (`sum(missing AD)`), and are (almost) all filtered out.

One de novo call survived, the WT parents must have had a populated AD value at that variant?

## Proposed Changes

  - By default, apply a GQ filter to all participants. Config entry `apply_min_all_sample_gq` to control this
  - If DP is in the schema, but DP is not defined for a sample, add a default value
    - Else if AD is in the schema, DP = sum(AD). If AD is in the schema but has a missing value, DP = default value.
      - If neither DP or AD are in the schema, give a default value for DP, but complain in the logs

End result - filtering process mostly unchanged, but instead of the filter accidentally throwing all the HomRef entries away by trying to use a field not defined in the schema, this will rely on `DP` if present, falling back to `AD`, then double falling back to a dummy default value which will pass filters.

## Checklist

- [ ] ShameLog Updated
- [x] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass

### DIFF
--- a/src/talos/example_config.toml
+++ b/src/talos/example_config.toml
@@ -120,10 +120,8 @@ min_depth = 5
 max_depth = 1000
 min_proband_gq = 25
 min_alt_depth = 5
-
-# un-comment this to apply a minimum GQ to all samples in the trio, including WT. This will reduce de Novo prediciton
-# if data was merged from single-sample VCFs (WTs/absent from single sample VCF will be inserted as WT/GQ=0)
-#min_all_sample_gq = 15
+min_all_sample_gq = 19
+apply_min_all_sample_gq = true
 
 [CreateTalosHTML.hyperlinks]
 template = "e.g. https://seqr.populationgenomics.org.au/COHORT_project_id/.../{sample}"


MattWellie:
Closed in favour of #630

PR #650, opened 2026-03-27
Cc nf corrections

MattWellie:
# Fixes

  - Various issues identified through analysis with claude. 
  - Some are soft inconsistencies (mixed use of case, missing shebangs, missing `nextflow.enable.dsl=2` etc.)
  - Some are hard bugs (bcftools output written using wrong flag)
  - Some are incomplete changes (mix of workflow publishing and output directives, addressed by unmerged https://github.com/populationgenomics/talos/pull/645)
  - Some are functional but incorrect (Memory setting of `"1 GB"` instead of `1.GB` - same result, but though string parsing instead of plain Groovy
  - Some are ideal but not essential (`set -euo pipefail` added to all script blocks), `rm -f` to prevent failure if files (which are definitely written) stop being written
  - Some corrections (indexing files during correct stage and passing VCF with index, instead of indexing an input, so the index would be lost in the working directory)
  - Adds missing `script:` directives to some files

## Proposed Changes

  - Solve all of the above. Test in progress
  -
  -

## Checklist

- [ ] ChangeLog Updated
- [ ] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ The input TSV uses two columns to locate variant input; `path` and `type`. `path
 2. **shards** a directory of pre-sharded multisample VCF fragments, each shard containing all samples.
 3. **ss_vcf_dir** single-sample VCFs, to be merged in the workflow, then sharded. These are detected using a glob, with the file extension controlled by `params.input_vcf_extension` (defaults to "vcf.bgz")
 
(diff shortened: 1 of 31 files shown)


MattWellie:
Closing this - all the same changes (and more) are in #645

PR #655, opened 2026-04-29
left-aligns column sorting, describe pseudodominant label

MattWellie:
# Fixes

  - the Pseudo-dominant MOI model and corresponding Flag weren't described in the report header
  - the column sorting arrows were right-aligned, and appeared pretty far from the column title they sort

## Proposed Changes

  * added new line to help box for pseudo label
  * left-align sorting arrow

Screenshot 2026-04-29 at 1 01 15 pm
Screenshot 2026-04-29 at 1 01 18 pm

### DIFF
--- a/src/talos/templates/index.html.jinja
+++ b/src/talos/templates/index.html.jinja
@@ -118,7 +118,7 @@
                             <span class="badge text-bg-warning">Ambiguous Cat.1 MOI</span> applied to AD Talos candidates that are P/LP in ClinVar but the gene is associated with both dominant and recessive disorders inheritance in PanelApp.
                         </p>
                         <p class="mb-1">
-                            <span class="badge text-bg-warning">Affected female with heterozygous variant in XLR gene</span> for females, we evaluate biallelic genes under a pseudo-dominant model, to detect plausible variants in the event of healthy-allele inactivation .
+                            <span class="badge text-bg-warning">Affected female with heterozygous variant in XLR gene</span> for females, we evaluate biallelic genes under a pseudo-dominant model, to detect plausible variants in the event of skewed x-inactivation.
                         </p>
 
                     </div>
--- a/src/talos/templates/index.html.jinja
+++ b/src/talos/templates/index.html.jinja
@@ -19,13 +19,9 @@
     <script src="https://cdn.datatables.net/rowgroup/1.0.2/js/dataTables.rowGroup.min.js"></script>
     <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/js/bootstrap.bundle.min.js"></script>
 
-    <style type="text/css">
+    <style>
         a {
             text-decoration: none;
-        }
-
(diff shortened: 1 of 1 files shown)


src/talos/templates/index.html.jinja
```diff
@@ -107,7 +115,10 @@
                             <span class="badge text-bg-warning">Low Read Depth</span> read depth (AD) for the sample is below the minimum threshold (default: 10)
                         </p>
                         <p class="mb-1">
-                            <span class="badge text-bg-warning">Ambiguous MOI</span> applied to AD Talos candidates that are P/LP in ClinVar but the gene is associated with both dominant and recessive disorders inheritance in PanelApp.
+                            <span class="badge text-bg-warning">Ambiguous Cat.1 MOI</span> applied to AD Talos candidates that are P/LP in ClinVar but the gene is associated with both dominant and recessive disorders inheritance in PanelApp.
+                        </p>
+                        <p class="mb-1">
+                            <span class="badge text-bg-warning">Affected female with heterozygous variant in XLR gene</span> for females, we evaluate biallelic genes under a pseudo-dominant model, to detect plausible variants in the event of healthy-allele inactivation .
                         </p>
```
SamBryen, line 122: I think the term 'healthy-allele' is misleading. Usually skewed x-inactivation occurs because there is something on that allele that is so pathogenic that it needs to be switched off for cell survival, such that over time you end up with skewed x-inactivation in some tissues. Presumably the variant under consideration in a pseudo-dominant model would always be less pathogenic than the switched off allele. 

Some suggestions:
"to detect plausible variants in the event of skewed x-inactivation"
"to detect plausible variants in the event of inactivation of the allele in trans."
"to detect plausible variants in the event of trans-allele inactivation" (not sure if this is good english)

MattWellie: I've gone with your suggestion: `for females, we evaluate biallelic genes under a pseudo-dominant model, to detect plausible variants in the event of skewed x-inactivation.`

MattWellie: My understanding was that one copy of X is epigenetically silenced permanently, so it's independent of the alleles? I guess having an active pathogenic allele could skew cell survival, but it wouldn't be correct to assume the silenced allele is always going to be the pathogenic one

cassimons (approved):
LGTM with Sam's suggestions

src/talos/templates/index.html.jinja
SamBryen: Yes one copy is epigenetically silenced permanently and it should be a coin flip. However you can sometimes see there is skewing in carriers of XLR diseases, especially for neurological disorders impacting the brain. For these genes, females are asymptomatic for the neurological disorder, but in the hypothetical situation where on the other allele they are mostly expressing they also carry, lets say a red/green colourblindness, then they will be colourblind even though this is usually recessive.

At least I think that's how that works. Does that sound right @rociorius?

PR #656, opened 2026-05-01
Extend mito annotations

### LINKED ISSUE #640: Handle optional files better
https://github.com/nf-core/sarek/blob/3.8.0/main.nf#L312

```
params.known_indels_vqsr ? Channel.value(params.known_indels_vqsr) : Channel.empty(),
```

and 

```
def sites_command = known_sites.collect { "--known-sites ${it}" }.join(' ')
```

MattWellie:
# Content

This PR is now an aggregation of the following changes. I stacked the changes, then consolidated them back down, so disentangling which change was originally each PR is a mess. I'll do better next time.

 - #663 - allows customisation of the AlphaMissense threshold
 - #661 - a painless panelapp parsing bug, no impact, but made generation of test fixtures difficult
 - #659 - adds a mkdocs page 

# Fixes

Mainly captured in https://cpg-populationanalysis.atlassian.net/browse/RD-944

  - Mitochondrial processing has not been present in the NextFlow workflow, and was in the private CPG implementation
  - Currently we only evaluate Mitochondrial variants using ClinVar, and don't apply any mito-specific annotation tools
  - BCFtools has been missing some clinically-relevant non-coding gene annotations
  - Closes #640

## Proposed Changes

  - downloads, reformats, and applies mitochondrial annotations from mitimpact, mitotip, and nAPOGEE (though these are not part of a `category` at this time, just available as annotations)
  - edits the default ensembl GFF3 file, renaming `MT` to `M` as a contig name. Talos is GRCh38-only, and by default the Ensembl GFF3 file uses a different naming convention to Hail, so the `bcftools csq` annotation is not compatible. Internally we do this [here](https://github.com/populationgenomics/references/blob/main/reference_generating_scripts/generate_bed_from_ensembl_manager.sh#L22).
  - adds the full mitochondrial analysis workflow loop into Nextflow, including annotation, labelling/selection, and feed-in to the MOI testing phase. This is fully optional, with a dummy file available for users which don't have/need mitochondrial analysis
  - Dockerfile builds a private fork of BCFtools containing a changed `bcftools csq` behaviour. This has been suggested upstream, with details here: https://github.com/samtools/bcftools/issues/2548
  - Test fixtures have been updated to contain [this variant](https://www.ncbi.nlm.nih.gov/clinvar/variation/3383301/), and is used to test the new bcftools change
  - Adds a new test data generation script to cook up 3 ClinVar pathogenic Mito variants, and adds the VCF into one of the test families. Test run demonstrates works as expected.

## What the heck, why are there so many changes? 

  - I'm using [echtvar](https://github.com/brentp/echtvar) as a fast annotator (gnomAD, AlphaMissense, now 3 mito datasets), but echtvar only knows how to process VCFs into its private `.zip` reference format, so we need to make spec-compliant VCFs out of the TSV/TXT files. To mediate I've got 4 parsing scripts, one for each of [AlphaMissense | Mito]. I've got 4 corresponding VCF header files too, and 4 echtvar config JSON files to tell it the data type for each of the different annotation fields. There was already an AlphaMissense VCF header file, but I've migrated all header files to a specific folder. A single Mito annotation is possible, but the 3 different input sources are versioned independently, so I'm keeping them separate
  - I'm adding Mitochondrial VCFs to the input spec. This is a new column in the input TSV (needs a note in the README to explain), but also a new field to adjust for every time the nextflow input channel is read or passed around.
  - There are changes to the preparation workflow (pre-processing all the Mito annotations), the AlphaMissense workflow (consolidating from 2 workflow files into one), the nextflow config (raw and processed names for all mito files), and new file channels are now passed into the `Talos.nf` workflow

## CI changes

  - limits some workflows to only running on pull requests, and only running when those pull requests are no longer in draft
  - limits workflow concurrency (e.g. in the docker building workflow). In the event of multiple quick commits, the most recent will cancel a now-redundant ongoing image build.

## Checklist

- [ ] ChangeLog Updated
- [x] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass


nextflow/talos.nf
```diff
@@ -80,12 +85,39 @@ workflow TALOS {
...
+    ch_mito_joined = ch_mts
+        .join(UnifiedPanelAppParser.out)
+        .join(StartupChecks.out)
+        .map { cohort, _mts, pedigree, config, _history, _ext, _seqr, mito, panelapp_data, check_file ->
+          tuple(cohort, mito, panelapp_data, pedigree, config)
+    }
+
+    ch_mito_branched = ch_mito_joined.branch {
+        real:     it[1].name != 'NO_MITO'
+        sentinel: true
```
MattWellie:
```suggestion
        sentinel: it[1].name == 'NO_MITO'
```

src/talos/annotation_scripts/parse_mitotip.py
```diff
@@ -0,0 +1,47 @@
...
+                continue
+
+            # pull out the relevant fields to bake into a VCF
+            position = tsv_line['Position']
+            ref = tsv_line['rCRS']
+            score = tsv_line['MitoTIP_Score']
+            mm_status = tsv_line['Mitomap_Status']
+
+            # write a VCF-format row
+            out.write(f'chrM\t{position}\t{ref}\t.\t{alt}\t60\tPASS\tmitotip={score};mitomap_status={mm_status}\n')
```
cassimons: Heads up — think the columns are slightly off here. Tab-splitting gives `chrM | position | ref | . | alt | …`, which lands `ref` in the ID column and `.` in REF. Compare to `parse_mitimpact.py` / `parse_napogee.py` which write `…\t{pos}\t.\t{ref}\t{alt}\t…`. Since `.` isn't a valid REF, `noodles-vcf` (echtvar) will likely reject at encode time, so `EncodeMitotip` would fail to produce the zip. Looks like a quick swap of the third and fourth fields.

echtvar/napogee_config.json
```diff
@@ -0,0 +1,6 @@
+[
+    {"field": "clingen", "alias": "clingen", "missing_string": "."},
+    {"field": "napogee", "alias": "napogee", "missing_string": "."},
+    {"field": "napogee_score", "alias": "napogee_score", "missing_value": -2147483648, "multiplier": 2000000},
+    {"field": "napogee_prob", "alias": "napogee_prob", "missing_value": -2147483648, "multiplier": 2000000},
```
cassimons, line 5: Tiny one — there's a trailing comma after this last array entry, which makes the file invalid JSON (`json.load` rejects it, and `serde_json` in echtvar is strict by default). `echtvar encode` would bail before even reading the VCF. Just dropping the comma should sort it.

src/talos/cpg_internal_scripts/cpgflow_jobs/annotate_mito.py
```diff
@@ -50,11 +50,27 @@ def make_bcftools_mito_job(
...
 
+    # mito annotation sources formatted for echtvar
+    mitimpact = batch_instance.read_input(config.config_retrieve(['references', 'mitimpact']))
+    mitotip = batch_instance.read_input(config.config_retrieve(['references', 'mitotip']))
+    napogee = batch_instance.read_input(config.config_retrieve(['references', 'napogee']))
+
+    csq_job.command(
+        f"""
+        echtvar anno \
+            -e ${napogee} \
```
cassimons: Just flagging — these are inside a Python f-string, so `${napogee}` renders as `$` followed by the resource path (e.g. `$/io/batch/.../napogee.zip`), which bash won't be able to use. The first `csq_job.command(...)` block above uses `{fasta}` / `{gff3_file}` without the `$`, which is the form that works. Same applies to the `${mitimpact}` and `${mitotip}` lines just below.

cassimons:
One general thing worth raising before merge — the change introduces a few silent breaks for existing users on the default config, and I'm not sure whether the intent is to handle them in code or document an upgrade path:

1. **Existing input TSVs lose a column.** `main.nf` does `file(row.mito, checkIfExists: true)` — for any cohort TSV without a `mito` column, `row.mito` is `null` and the workflow aborts at startup with a fairly opaque error. Users also need to know that `nextflow/assets/NO_MITO` is the specific sentinel filename (`talos.nf:97` branches on `it[1].name == 'NO_MITO'`, so other empty placeholders won't work).

2. **`params.ensembl_gff` default flips** to the new `MTtoM`-renamed file. Anyone on the default config will hit `checkIfExists: true` at startup until they re-run `gather_file.sh` (which does the rename in place from the existing download — lightweight, but they have to know).

3. **`talos.nf` requires the three new mito echtvar zips to exist at startup, even for cohorts with no mito data**, because the `checkIfExists: true` calls on `mitimpact_zip` / `mitotip_zip` / `napogee_zip` happen before the `NO_MITO` branch. A user upgrading for unrelated reasons and never running mito will still be forced to re-run `preparation.nf`. This one feels like it could be guarded (only check the zips when at least one cohort has real mito data) — happy to do that as a follow-up if you'd like, or leave it as a documented upgrade step.

Question: do you want to soften any of these in code (e.g. defaulting `row.mito` to `NO_MITO`, deferring the zip existence checks to inside the branch), or just land an upgrade-notes section in the README + a 10.0.4 changelog entry walking through the four steps (TSV, gather_file.sh, preparation.nf, docker rebuild)? Either is fine, just want to make sure it's deliberate.

cassimons:
☝️ obviouly all from CC - but the main point stands, what is the upgrade path here?

src/talos/cpg_internal_scripts/cpgflow_jobs/annotate_mito.py
MattWellie: oops, that's a rubbish copy-paste error from the nextflow script block

echtvar/napogee_config.json
MattWellie: Aha! Not this time chatbot!

Echtvar uses [JSON5](https://json5.org/) which is an extended specification: "Arrays may have a single trailing comma."

src/talos/annotation_scripts/parse_mitotip.py
MattWellie: not sure how that happened, but yup, it was writing duff VCFs for sure

MattWellie:
@cassimons all fair points - 

The plan is to keep loading new things into the `dev` branch, and then create an explanatory changelog/new features entry on the docs page/readme/changelog/release tag/other when it's released to main. However... makes sense for the workflow to be flexible to inclusion/absence of all optional inputs.

The current TSV input contains a number of optional columns (history file, secondary IDs, seqr IDs, mito data), so I've changed the parser to allow for those columns to be missing, in which case the absent files are replaced with the expected defaults. I've run the test workflow with all columns present/all optional columns missing, and checked the parsing is robust to both

I've made the change to if/when the mito resource files are read in, so they'll only be required files when the mito workflow is being used.

The chrM adjusted GFF file I'm less sure about... I think it's reasonable to prompt users to run the preparation workflow at least once with each release, especially as it contains new functionality which requires new files. We want them to be rerunning on a regular basis, so this would be useful to encourage. I'll make sure that it's in the changelog/release notes, but failing to use the edited chrM gff would mean the mito workflow is broken, whilst a gff with the change will work with or without the mito wf.

cassimons:
Sounds good.
> I think it's reasonable to prompt users to run the preparation workflow at least once with each release
Absolutely, but you might need to add a sentinel that will let it throw a warning each run until they run the update

MattWellie:
I think that's solved for here, if only by accident 😬 

- The old and new (with mito) nextflow.config both reference `params.ensembl_gff`
- The latest file download script will download the GFF, then produce a second version with "MT" -> "M", but the original version is kept
- If it's run using an old config and no mito, it will use the old GFF file, which still exists
- If it's run using new code and old config, but no mito data, we'll never try to load in mito-specific resources, and the old GFF is fine
- If it's run using the old config and mito it will break, prompting user to download and prepare fresh resources
- if it's run using a new config and old workflow (?), the ensembl GFF will exist just fine, and is interchangeable


(the mito data, reference material & input TSV column, only being loaded when required is key to this flexibility, and was a result of the suggestions in this review. The original version was pretty inflexible and required all existing users to update their input files & run a prep workflow)

PR #663, opened 2026-05-15
Custom alphamissense threshold

MattWellie:
# Fixes

  - Currently the alphamissense category is fixed at `am_class.contains('pathogenic')`, which is AlphaMissense's canonical decision about the pathogenic score threshold (0.564). This is too lenient for some users, which is part reason it's always been a support-only category
  - Thread: https://centrepopgen.slack.com/archives/C037V16B41Y/p1778624785272479

## Proposed Changes

  - The AlphaMissense category decision is now based on a float value, customisable in config. This solves for the use case of "we'd want AM to be a full category, but only with a stricter threshold"
  - to clarify, with the default config this will assign the AM category at the same net score as before, and AM will still be a support category, but both attributes (strictness and support/full) are independently modifiable per-user

## Unrelated
  - Updates uv.lock for the mkdocs libraries, should be in the previous PR
  - The method which tries to harmonise Category names doesn't currently account for Spaces or punctuation. i.e. if a variant was previously `ClinVarP/LP` (no space), and now it's formatted as `ClinVar P/LP` (space separated), the two don't match, and both show up in the report. This looks dumb. Now the category-name matching method first flattens the string, removing punctuation and spaces, then lowercasing the result. This is looked up to find the better formatted name to present, which should prevent double-display.

## Checklist

- [x] ChangeLog Updated
- [x] Version Bumped
- [x] Related Issue created
- [x] Linting checks pass


docs/Configuration.md
```diff
@@ -25,13 +25,14 @@ This document explains the configuration options, the default values, and sensib
...
+| `callset_af_sv_recessive` | A coarse population and callset frequency filter - if an SV variant is more common than this either within the callset, or in the gnomAD annotations, remove                                                                                               | 0.03                                                                          |                                                                                                   |
+| `am_pathogenicity`        | A float, defining the minimum AlphaMissense pathogenicity score to be treated as pathogenic. Replaces deference to AlphaMissense's default am_class value, which is using the 0.564 score as a Likely Pathogenic threshold                                 | 0.546                                                                         |                                                                                                   |
```
SamBryen: Value should be 0.564 not 0.546, it is correct in the description but not the value set

nextflow/inputs/config.toml
```diff
@@ -36,6 +36,7 @@ minimum_depth = 10
 spliceai = 0.5
 csq_string = [ "consequence", "gene_id", "gene", "transcript", "mane_id", "mane", "biotype", "dna_change", "amino_acid_change", "codon", "ensp", "am_class", "am_pathogenicity",]
```
SamBryen, line 37: If 'am_pathogenicity' is replacing 'am_class', does it need to be remove here?

test/test_hail_categories.py
```diff
@@ -67,21 +67,21 @@ def test_spliceai(splice_score, classified, make_a_mt):
...
+    'am_pathogenicity,classified',
     [
-        ('likely_pathogenic', 1),
-        ('not_pathogenic', 0),
-        ('', 0),
-        (hl.missing('tstr'), 0),
+        (1, 1),
+        (0, 0),
+        (0.88, 1),
+        (hl.missing('tfloat64'), 0),
```
SamBryen, line 75: Questions for learning purposes: this is a test that checks that your logic is being handled correctly? Before this PR, am_class had three options 'likely_pathogenic' (ie. >0.564), 'not_pathogenic' (ie. <0.564) or null. You want to return the value of 1 if it's likely_pathogenic and 0 if not. Now that we don't have am_class, we are directly comparing the value in the config to the number 1 and 0 returning the 1 and 0 respectively. We're also testing a value of 0.88 that is above 0.564 and so we would expect a 1, is this value chosen at random? Won't this test fail if we set the config to 0.906?

SamBryen (approved):
Insofar as I can read looks good! Once you fix the typo on line 35 of docs/configuration.md I think it's good to go? Maybe check that test question first, probably me just not understanding how tests work.

docs/Configuration.md
MattWellie:
```suggestion
| `am_pathogenicity`        | A float, defining the minimum AlphaMissense pathogenicity score to be treated as pathogenic. Replaces deference to AlphaMissense's default am_class value, which is using the 0.564 score as a Likely Pathogenic threshold                                 | 0.564                                                                         |                                                                                                   |
```

MattWellie: good spot!

nextflow/inputs/config.toml
MattWellie: Hmm. I always took both, and the transcript table always shows both values e.g. `likely_pathogenic (0.85)`.

On the one hand we could drop the am_class as it's not the basis for assigning the category any more.

On the other, if left as is, the transcript table will present the AlphaMissense original score and class whether it's benign, unknown, or pathogenic. Is that still useful?

test/test_hail_categories.py
MattWellie: The test suite loads the config file present at `test/input/config.toml`, and applies the various tests WRT the values in that file. If the config was updated to 0.9 the test would fail, but it would be a bit self inflicted - to pass the config and test would have to be updated to work as a pair. 0.88 was a randomly chosen value, the test is >= the threshold, so anything above that will pass. A super exact test would use 0.563 as a fail, and 0.564 as a pass.

And yeah, the test runs 4 times, each one annotates a dummy variant, then passes it through the AM categorisation method. It used to annotate in class, now it annotates a numerical value instead. When the category is applied, it annotates `categoryalphamissense=1`, or 0 for a fail. The 0/1 expected is a check that when the results are collected, it's a single pass or fail.

nextflow/inputs/config.toml
SamBryen: Oh right, these are the values displayed, yes I think that is still useful to keep then. It still gets generated somewhere? That part hasn't been touched in this PR?

MattWellie: Yup yup, we always pulled class and pathogenicity, but were only making the category decision using class. There's a top-level value in the report for `Max AlphaMissense` which always displayed the highest number amongst transcripts, and the transcript table showed both

SamBryen (approved):

EddieLF (approved):
LGTM

PR #665, opened 2026-05-22
Stability tweaks

MattWellie:
# Fixes

  - Solves some current runtime issues with VMs running out of storage or memory, and some QOB-driver jobs being preempted, causing duplicate QOB workflows

## Proposed Changes

  - Makes the arguments for bcftools and echtvar job `storage`, `cpu`, and `memory` configurable
  - Sets the driver job for QOB tasks to be non-spot

n.b. this only affects CPG-methods


Johnnyassaf (approved):
Looks good to me!

PR #666, opened 2026-05-22
more QOB juice

MattWellie:
QOB jobs are chronically unstable at the moment, needs a resource bump to get the batches finished


Johnnyassaf (approved):
The workers yearn for more means of production, maybe make the cores configurable?
Also, version bump if you dont like the -2

PR #667, opened 2026-05-22
Includes STRs as a data source

MattWellie:
# Fixes

  - No STRs

## Proposed Changes

  - Adds an STR VCF as an optional variant source
  - Adds in a repeat disorders panel - only STRs which are pathogenic according to STRipy and PanelApp are considered
  - Creates a new ShortTandemRepeat variant type, which contains the per-sample genotypes and repeat count for each allele. It also records the diseases relevant to each locus and pathogenic count thresholds for presentation in the report
  - Makes a few pieces in ValidateMoi easier - i.e. instead of keeping track of the small variant and (optional) mito, sv, and now STR VCFs, as separate objects which need to be passed around explicitly they're now loaded into a dictionary which looks like:

```
vcf_sources = {
    'small': small_variant_vcf,
    'sv': sv_vcf | None,
    'str': str_vcf | None,
    'mito': mito_vcf | None,
}
```
  - adds a new child-row block specific to STRs, which shows each disease relevant to the locus, its ranges and thresholds, and a RAG rating for whether the STR detected exceeds the pathogenic threshold for the specific disease

## Checklist

- [x] ChangeLog Updated
- [x] Version Bumped
- [ ] Related Issue created
- [x] Linting checks pass


MattWellie:
n.b. I think the wiring here all makes sense, but the exact VCF specification/INFO/FORMAT fields matching what we're populating internally is a bit shortsighted... ExpansionHunter VCFs would be a more generic target.

MattWellie:
I'm re-opening this. It's tightly scoped to a very specific STR caller, but it's a good chunk of effort and better than nothing. Improving (e.g. targetting any-old STR caller and its catalog) can come later if there's appetite outside CPG/VCGS/Broad.

It also includes a number of tidy-up pieces like how multiple variant sources are handled, which I would like to keep

MattWellie:
memo to self - I need to ping off some test runs now that this has been rebased (messy)

PR #671, opened 2026-05-29
Never modify the global config during run setup

MattWellie:
# Fixes

  - The CPG-internal implementation is mixing up config settings across projects
  - Root cause - blocks of config were being referenced, but also `update`d within a loop. e.g. the `GeneratePanelData` part of config was being `updated` to include a specific project's forced_panels. 
  - If the next project didn't have any forced panels, the previous project's panels would be inherited due to the base config having been updated.

## Proposed Changes

  - use copy - each project pulls it's part of the config as a separate dictionary
  - if the project has specific updates to make (e.g. forced panels), those changes are made to a private copy of the config, instead of potentially affecting all projects processed after

Chocolate fish for @SamBryen 🐟

### DIFF
--- a/src/talos/cpg_internal_scripts/cpgflow_jobs/make_config.py
+++ b/src/talos/cpg_internal_scripts/cpgflow_jobs/make_config.py
@@ -1,3 +1,4 @@
+import copy
 import functools
 import json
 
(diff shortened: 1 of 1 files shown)


src/talos/cpg_internal_scripts/cpgflow_jobs/make_config.py
SamBryen, line 103: Gemini explained copy.deepcopy() to me and I think this makes sense. Though why would you not use deepcopy for sequencing type as well? Could you accidently run a genome as an exome and vice versa?

MattWellie: 2 reasons here -
- sequencing_type is just a String, and is only read instead of read + updated. The issues here were because the config blocks that are now being `copied` are a block of generic settings, and those are updated with cohort-specific settings. i.e. `GeneratePanelData` is a block of config that applies to every run, and that would have cohort-specific content (panels) added into it. That update was done badly so cohort-specific content leaked into the non-specific content
- exomes and genomes (and long reads) are run separately, so there's never an instance of exome settings leaking over to a genome run

SamBryen: Ah that makes sense then! Happy to approve, but looks like you've already merged it? How did you merge without at least 1 approval?

MattWellie: Ah I squashed it in with Talos admin permissions 🤐 

Also fun observation; in [this PR](https://github.com/populationgenomics/production-pipelines-configuration/pull/171) there was the text "code fix [here](https://github.com/populationgenomics/talos/pull/671)". When that PR was merged, GitHub read "fix LINK" and tried to close this PR, but it hadn't been approved, so it just scrapped it 😅.

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

PR #687, opened 2026-07-27
CPG-internal: add improved LRS discovery

MattWellie:
# Fixes

  - In short-read sequencing we have a consistent 1:1 relationship between the Cohort incorporated into the callset, and the Cohort being used in Talos, so dataset discovery is simple
  - In LRS we have Cohorts stratified by multiple values (platform, amplification/library), so there can be multiple live Cohorts for a dataset.

## Proposed Changes

  - Build on the existing Cohort-namespacing, so that multiple runs in the same dataset won't clash, e.g. `gs://cpg-PROJECT-main-tmp/long_read/talos/COHORT/ExtractVcfFromMt/COHORT_separate.vcf.bgz`
  - Add new method for correlating a LRS Cohort with the available LRS MT options
  - Use the `analysis.sequencingGroups` attribute combined with the `cohort.sequencing_group_ids` to check for commonality
    - All for 'any' and 'exact' analysis matching profiles, deciding how strict to be


MattWellie:
This specific change (LRS discovery) requires a change - we're going to register the sequencing technology in the long-read meta, instead of having to detect it implicitly from elements of the file path. Once analysis record contents have been updated, this can be changed

PR #691, opened 2026-08-06
Adapt to improved meta, consolidate discovery methods

MattWellie:
# Fixes

  - long read analysis entry discovery was being done implicitly, through identification of `long_read` or `LongRead` in the file paths. We may not retain this convention

## Proposed Changes

  - analysis entries relating to long-read are now tagged using the convention `sequencing-technology=long-read` in the meta block of the analysis. This is now fed into the metamist queries where appropriate
  - I'd retained two methods, one for the MT records we have `analysis.outputs` as a String, and one for the records we will have once https://github.com/populationgenomics/metamist/pull/1306 is merged (`analysis.outputs.path` as a String). Instead of maintaining two methods, there is now an `isinstance` check done on the returned objects, so one method handles both cases

n.b. due to the deletion of a whole method, the diff here looks a little misleading. The file was previously:

  - general method
  - future LRS method
  - current LRS method

and is now

  - general method (with a block of indentation removed)
  - current LRS method

### DIFF
--- a/src/talos/cpg_internal_scripts/cpg_flow_utils.py
+++ b/src/talos/cpg_internal_scripts/cpg_flow_utils.py
@@ -196,11 +196,15 @@ def query_for_latest_analysis(
     analysis_by_date = {}
     for analysis in result['project']['analyses']:
         outputs_block = analysis['outputs']
+        output_path: str | None = None
         if isinstance(outputs_block, dict):
-            output_path = outputs_block['path']
+            if (opath := outputs_block.get('path')) and isinstance(opath, str):
+                output_path = opath
+            elif (opath := outputs_block.get('output')) and isinstance(opath, list):
+                output_path = opath[0]
         elif isinstance(outputs_block, str):
             output_path = outputs_block
-        else:
+        if output_path is None:
             loguru.logger.debug(f'Skipping {outputs_block} in dataset {query_dataset}, unexpected type.')
             continue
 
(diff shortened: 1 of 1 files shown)


src/talos/cpg_internal_scripts/cpg_flow_utils.py
```diff
@@ -316,11 +266,13 @@ def query_for_latest_lrs_mt(
...
     analysis_by_date = {}
     for analysis in result['project']['analyses']:
-        output_path = analysis['output']
-
-        # manually implementing an XOR check - long read (bool) and LongRead in output must match
-        if LONG_READ_STRING not in output_path:
-            loguru.logger.debug(f'Skipping {output_path} in dataset {query_dataset} - not long read.')
+        outputs_block = analysis['outputs']
+        if isinstance(outputs_block, dict):
+            output_path = outputs_block['path']
```
EddieLF:
```suggestion
            if 'path' in outputs_block and isinstance(outputs_block.get('path'), str):
              output_path = outputs_block['path']
            elif 'output' in outputs_block and isinstance(outputs_block.get('output'), list):
              output_path = outputs_block['output'][0]
```

Unfortunately some analyses have been written like so:
```
{
"id": 413891,
"outputs": {
  "output": [
    "gs://cpg-dataset-main/mt/LongReadSNPsIndels-hash.mt"
  ],
  "secondary_files": {
    "inputs": {
      "id": 313632,
      "parent_id": 313647,
      "path": "gs://cpg-dataset-main/lrs_annotation/snps_indels_annotation/hash/input_vcfs.json",
      ...
    }
  }
}
```

MattWellie: so many variations 😢

PR #714, opened 2026-09-11
Confidence level tracked in history

MattWellie:
# Fixes

  - Twinned with https://github.com/populationgenomics/talos2/pull/10
  - Confidence increase (e.g. Amber -> Green) wasn't recorded in the "date evidence last changed" attribute

## Proposed Changes

  - see better explanation on https://github.com/populationgenomics/talos2/pull/10

## N.b.

This is probably a logical stopping point before migrating Talos2 into this repo

## Checklist

- [ ] ChangeLog Updated
- [ ] Version Bumped
- [ ] Related Issue created
- [ ] Linting checks pass


src/talos/templates/index.html.jinja
```diff
@@ -177,7 +177,7 @@
                             <th>Variant</th>
                             <th>Gene (MOI)</th>
                             <th>Categories</th>
-                            <th>First Tagged</th>
+                            <th>Evidence Updated</th>
```
cassimons, line 180: amber>green is a no brainer - but I feel like we should probably pause on this and make sure we have a clear philosophy on how we want to track and expose changes in evidence (of different types). We have had a couple of half-attempts at this over time, but before we change it again we should probably try to come up with a more holistic and understandable philosophy for what changes we track and expose.

MattWellie: @cassimons I've had another whack at this - we've been tracking a lot of changes, but only keeping them in the results JSON file and never in the report. The intention was to keep the report clear, keep the JSON rich, and prioritise the earliest date to make incremental report triage as quick as possible. The things we've been tracking are:

1. category & date of assignment
2. date of first phenotype match, and if a variant is newly phenotype matched
4. clinvar stars & if the star count has increased since previous run
5. support variants this has been paired with as a comp-het
6. in this PR - confidence of gene in panelapp, and if it has increased since previous run

All these have contributed to a timestamp for 'evidence last updated', but we've not displayed it. 

Looking at these things again - 

1. New categories assigned is useful. It's probably not common, but relevant to interpretation. These will mostly be new in ClinVar, or Clinvar-0-star to a full Clinvar P/LP
2. Date of first phenotype match is not too exciting, this is reasonably static
3. Clinvar star increase - same as point 1, this will mostly appear as a new category, unless it's P -> expert review, which will be very low numbers
4. Support variants - the newly classified supporting variant would have today's date, so attention is already drawn to this
5. PanelApp confidence - a rise to Green is important, Red->Amber less so. Choosing to run Talos with Amber or Red genes is a per-site decision, and probably indicative of how far users want to look

So... Of all the things I've been tracking, the first tagged is most important, optionally with a bit of emphasis to show when a gene becomes green, an additional category is assigned, a variant is now a phenotype match, or ClinVar confidence increases. 

I've had a play around with using the report to show just the first date of discovery, with additional emojis to identify when these criteria are met. There's still only a single displayed date, but there's a slight emphasis in the display:

Screenshot 2026-09-21 at 7 14 46 am

```
REASON_EMOJIS = {
    'new': '🆕',      # new category assigned
    'clinvar': '💫',  # rise in clinvar stars compared to previous max
    'pheno': '🎯',   # newly a phenotype match
    'green': '🚦',    # newly promoted to green
}
```

Is this a stupid idea? Even if these emoji highlights are disabled, these will continue to be logged in the JSON, but evidence last updated will no longer consider new partner variants, or a rise in clinvar stars, because we already track that in other ways.

